Issue #30, opened 2018-09-09
seaborn fails


yannikschaelte:
The seaborn.PairGrid, as used in pyabc.visualization.plot_kde_matrix, seems to have changed its api. namely, one runs into ``TypeError:  cannot concatenate object of type "class 'numpy.ndarray'>"; only pd.Series, pd.DataFrame, and pd.Panel (deprecated) objs are valid``, triggered by pyabc.visualization.py line 310 ``df = pd.concat((x,), axis=1)``. Reason seems to be that seaborn since version v0.9.0 (July 2018) converts an array to numpy before passing it on to the plotting function (https://github.com/mwaskom/seaborn/blob/master/seaborn/axisgrid.py#L1390). 

Suggestion: We need the column name, so check whether this is passed somewhere else, or just drop seaborn.

Issue #29, opened 2018-09-09
Improve chemical_reaction.ipynb


yannikschaelte:
In the notebook, there are some minor inaccuracies:
* the description and code implementation of the distance are normalized differently
* the ``Y`` component is observed, not ``X``

Issue #28, opened 2018-09-09
Fix FutureWarning


yannikschaelte:
Following warning occurs on running pyabc:
```
FutureWarning: Method .as_matrix will be removed in a future version. Use .values instead.
```
in:
```
pyabc/storage/history.py:200
pyabc/transition/multivariatenormal.py:64
pyabc/smc.py:729
```
among others.

PR #36, opened 2018-09-10
Fix various issues

### LINKED ISSUE #34: plot_kde_matrix: wrong axis labels for diagonal
For the diagonal, the y axis label is wrong bc taken as the range of the y parameter.

yannikschaelte:
* R installation on travis failed due to expired keys or something like that. Now specified more precisely.
* Fix some descriptions in chemical_reaction.ipynb, see #29.
* No longer use seaborn for plot_kde_matrix, because they changed their api so that it no longer works that easily. Instead just assemble the plot oneself, looks almost the same, see #30  (and on the way also fixed #34)
* Some adaptations bc used packages changed, see #28 .


elbaraim:
* New visualization routine: Disagreement between Y axis of the subplots top-left and bottom-right. The old version show values up to one (I guess this is normalized?), while the new version not. Is there any reason for that to be changed?

* R installation on travis failed due to expired keys or something like that. Now specified more precisely. --> Where can I find the commit corresponding to this ?

yannikschaelte:
>     * New visualization routine: Disagreement between Y axis of the subplots top-left and bottom-right. The old version show values up to one (I guess this is normalized?), while the new version not. Is there any reason for that to be changed?
yes, beforehand it was afaik the parameter limits on all axes, which for the diagonal plots makes no sense (look e.g. here https://pyabc.readthedocs.io/en/latest/examples/multiscale_agent_based.html). instead, now it is the true range of the posterior density approximation shown here. one could normalize this, but i would rather not do that.
> 
>     * R installation on travis failed due to expired keys or something like that. Now specified more precisely. --> Where can I find the commit corresponding to this ?
yeah, that took a few commits, because I could only test it once uploaded. it's commits 3 ("try not insist on trusty") to 9 ("add comments") above. as far as i understood, it was finally that some certificate had expired, and I found instructions in various forums that one should do it as I did now in .travis.yml. hope it will continue to work.

PR #43, opened 2018-11-28
Fix pickling

yannikschaelte:
reduce the amount of data that has to be pickled for multiple of the employed servers. previously, the maximum memory usage was linear in

+ the summary statistics size,
+ number of accepted particles, 
+ and the number of sampling workers. 

the reason was that when pickling the sampling function, also (parts of) the full ABCSMC class was pickled, which contains the accepted particles of the last accepted generation.

now, it should be possible to use pyabc with hundreds of workers and large data.

also:
+ make tidying optional in abc-export
+ update primary visuals (logo, readme)

update version -> 0.9.3


yannikschaelte:
see also some analysis here: https://github.com/yannikschaelte/Div/tree/master/pyabc_tests/redis

yannikschaelte:
yes. good points, but for some reason flake8 will not allow to insert spaces here, only if there is a datatype in addition (e.g. ``n: int = 0``) The out_format was needed, because format is a built-in python function (which is also why one cannot use names like dir, file, etc.).

PR #44, opened 2018-12-18
Fix kde

yannikschaelte:
+ allow to pass kde in visualization routines
+ allow to re-submit observed sum stats to ABCSMC.load() function

Version update to 0.9.4.


yannikschaelte:
on the different scaling constant: good point. I thought it might make more sense to reduce it here to avoid over-smoothing in the visualization, and therefore be better able to detect multimodalities.
or would you leave it at the default of 1 to ensure the behavior does not change compared to previous versions?

elbaraim:
I would either:
- include a short line of text in `doc/releasenotes.rst` because users may execute the same code using the new pyABC version and get different results (in case they use default values). Maybe is good to point this out (?)
- keep it as it is

What do you think?

yannikschaelte:
I set it back to 1 now (because I actually never checked whether 0.5 is really better; usually only much smaller values show multimodalities). but we might want to keep this in mind. I added a warning to the visualization documentation that one must be aware of this.

the typo is solved. the title thing I would leave as it is now, because it gives more flexibility, but I have actually never used a title. Before, I think it was "Posterior" which is not too informative either (more like marginal posterior approximation).

PR #53, opened 2019-02-01
Fix rtd

yannikschaelte:
Fix several issues in rtd doc build.
* Remove unnecessary dependencies in conda and pip. This reduces the build time from 12 min to 5-8 min currently.
* In particular, this is possible bc if we run all the notebooks beforehand (done now), this does not need to be done on the server.
* Tidy up build requirements for travis, and for rtd.
* Change html template to the rtd one, since it has some preferable features (in particular better design and better linking structure).
* Add links to all rst pages so that the correct toc is always displayed.
* Several minor fixes.

Further:
* Add long description to setup.py for pypi


yannikschaelte:
To see how the doc looks after the updates, see here: https://pyabc.readthedocs.io/en/fix_rtd/
The link will be only available during this pull request though.

PR #63, opened 2019-02-20
Develop

yannikschaelte:


pyabc/population.py
```diff
@@ -196,8 +196,8 @@ def get_weighted_distances(self) -> pandas.DataFrame:
         Returns
         -------
 
-        weighted_distances: pandas.DataFrame:
-            A pandas.DataFrame containing in column 'distance' the distances
+        weighted_distances: pd.DataFrame:
+            A pd.DataFrame containing in column 'distance' the distances
```
elbaraim: I would write "A pandas DataFrame ... "
The abbreviation of `pd` is widely used, but it is not the exact name of the package

pyabc/storage/history.py
```diff
@@ -123,64 +143,64 @@ def all_runs(self):
     @with_session
     def _pre_calculate_id(self):
         abcs = self._session.query(ABCSMC).all()
-        if len(abcs) == 1:
-            return abcs[0].id
+        if len(abcs) > 0:
```
elbaraim: why this change?

pyabc/storage/history.py
```diff
@@ -123,64 +143,64 @@ def all_runs(self):
...
         -------
 
         df, w: pandas.DataFrame, np.ndarray
-
-        df:
-            is a DataFrame of parameters
-        w:
-            are the weights associated with each parameter
+            * df: a DataFrame of parameters
+            * w:  are the weights associated with each parameter
```
elbaraim: two empty spaces here ;)

pyabc/storage/history.py
```diff
@@ -758,21 +816,26 @@ def get_population_strategy(self):
...
         Parameters
         ----------
-        m: int, optional
-            The model to query.
-            If omitted, all models are returned
-        t: str, optional
-            Can be "last" or "all"
+
+        m: int or None, optional (default = None)
+            The model to query. If omitted, all models are returned
```
elbaraim: full spot at the end of the sentence :dancer:

pyabc/storage/history.py
```diff
@@ -758,21 +816,26 @@ def get_population_strategy(self):
...
+
+        t: int or str, optional (default = "last")
+            Can be "last" or "all", or a population index.
             In case of "all", all populations are returned.
-            If "last", only the last population is returned.
+            If "last", only the last population is returned, for an int value
+            only the corresponding population at that time index.
+
```
elbaraim: too picky: but the case of `population index` given is not explained :dancing_women:

pyabc/storage/history.py
yannikschaelte: what stood there beforehand means that only if there is just 1 analysis in the db, then the id is set to it.
now the id is always set to the lastest added analysis, cause that's what I would expect (it is the most relevant one).

yannikschaelte:
sorry, i had closed th pr already. these things are indeed true, I will add them in #56

pyabc/population.py
yannikschaelte: good point. I used pd here now cause I had changed the import to ``import pandas as pd``. Best would be to generate links to the respective classes here ...

PR #73, opened 2019-03-23
Feature viz

yannikschaelte:
Add more visualization routines.


yannikschaelte:
Visualization routines are now used in https://github.com/ICB-DCM/pyABC/blob/feature_viz/doc/examples/parameter_inference.ipynb.

Issue #114, opened 2019-05-28
Distance updates for AdaptivePNorm take a very long time when the fraction of accepted samples is small.


Pat-Laub:
Firstly, thanks very much for creating this great library!

When using AdaptivePNorm with a small `min_acceptance_rate` (< 1e-3), I noticed that sometimes pyABC spent more time running some of the update steps (i.e. some of the unparallelised steps) than it did running the parallel simulation steps. 

In particular, in smc.py at the line 
    `self.distance_function.update(t + 1, sample.all_sum_stats)`
it would spent a long time just evaluating the `sample.all_sum_stats`. This is because it would load all accepted and rejected summary statistics in order to calculate (for example) the sample standard deviations of each summary statistic without bias towards accepted particles.

This procedure is fine, however it seems that calculating the sample standard deviation with **all** of the particles can be a bit of overkill when the number of rejected particles gets into the hundreds of thousands or millions. I made some modification in a fork which allows one to update the distance using a smaller sample of particles than the entire history; I hope this is useful.

yannikschaelte:
Hi, thanks for your message and pull request. We so far never had the problem of overly many samples, but for small simulation times this can apparently happen, as in your case. Thanks for reporting this.

In such a situation, a "representative sample" of accepted and rejected summary statistics should indeed suffice. Thus, your pull request is most welcome and we will definitely include it ASAP.

Maybe we will make minor modifications, like allowing to alter the number of samples taken, basing the default on a multiple of the number of accepted particles, and making the limitation optional. One could also think about not memorizing samples at all as soon as a threshold, say 1000, has been hit, in the first place.

> I noticed that sometimes pyABC spent more time running some of the update steps (i.e. some of the unparallelised steps) than it did running the parallel simulation steps.

When you notice more things of this kind, please let us know. We are aware of some issues with e.g. database access, which we could then address.

yannikschaelte:
closed via #130

PR #115, opened 2019-05-28
Added option to limit number of particles for adaptive distance update.

Pat-Laub:
As per https://github.com/ICB-DCM/pyABC/issues/114

### DIFF
--- a/pyabc/sampler/base.py
+++ b/pyabc/sampler/base.py
@@ -34,6 +34,23 @@ class Sample:
         return sum((particle.accepted_sum_stats + particle.rejected_sum_stats
                     for particle in self._particles), [])
 
+    def first_n_sum_stats(self, n):
+        """
+        Get (at most) the first n summary statistics.
+
+        Returns
+        -------
+
+        n_sum_stats: List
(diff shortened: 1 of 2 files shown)


pyabc/sampler/base.py
```diff
@@ -34,6 +34,23 @@ def all_sum_stats(self):
         return sum((particle.accepted_sum_stats + particle.rejected_sum_stats
                     for particle in self._particles), [])
 
+    def first_n_sum_stats(self, n):
```
yannikschaelte: Note that a particle.accepted_sum_stats list can contain >1 entries. So the number might be >n here actually. But this is an old feature of pyabc which I am not aware of anyone using (and will probably be deleted soon), so I guess this should be fine.

yannikschaelte:
@Pat-Laub Please note that we will change the default number of records to inf in version 0.9.23. This is because in multiple applications we really need the full info (aside from distances, also some epsilons will employ the info soon), while using only e.g. 1000 would create too much of a bias, such that we consider a default of inf safer in general. One can however set it manually by the constructor argument `max_nr_recorded_particles`.

PR #116, opened 2019-05-29
fix some rpy2 problems

yannikschaelte:


yannikschaelte:
there are some unresolved issues left (failures with multithreading in R/rpy2) which cannot be addressed here, so this has to be stalled.
For singlecore execution, all seems to work fine locally. On travis, there is another problem with ../../../virtualenv/python3.6.3/lib/python3.6/site-packages/rpy2/rinterface_lib/openrlib.py:17: in _dlopen_rlib
    rlib = ffi.dlopen(lib_path)
E   OSError: cannot load library '/usr/lib/R/lib/libR.so': dlopen: cannot load any more object with static TLS
https://travis-ci.org/ICB-DCM/pyABC/builds/538638913?utm_source=github_status&utm_medium=notification

yannikschaelte:
Some updates: 

The dlopen error does not occur on ubuntu xenial. I suppose there's nothing bad about moving to a later version, so that should be fine.

Multicore execution still fails due to some unknown reason (maybe the R instance in rpy2 being a singleton is a problem?). Singlecore seems to work, on redis workers fail but this can be handled there. Whether things change needs to be checked at some points then.

Staying with rpy2 is no option, since it includes code that will discontinue support e.g. in pandas soon.

PR #117, opened 2019-05-30
Fix travis, R, sklearn

yannikschaelte:
- update travis version to xenial (16.04) since rpy2 causes problems otherwise with dlopen (see #116)
- fix argument change in sklearn GridSearch, update to 0.21.2

This is only a preliminary fix, using an outdated R link, because the latest version of rpy2 could not be be completely included yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,16 +5,13 @@ sudo: true
 os:
 - linux
 dist:
-# take the latest supported LTS version
-- trusty
+# take the latest supported version
+- xenial
 
 # install dependencies
 
 before_install:
-# install latest r
-- sudo sh -c 'echo "deb http://cran.rstudio.com/bin/linux/ubuntu trusty/" >> /etc/apt/sources.list'
-- gpg --keyserver keyserver.ubuntu.com --recv-key E084DAB9
-- gpg -a --export E084DAB9 | sudo apt-key add -
+# install r
 - sudo apt-get update
 - sudo apt-get install r-base r-base-dev r-base-core r-recommended
 # from apt
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,10 @@ dist:
 # install dependencies
 
 before_install:
-# install r
+# install latest r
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -6,13 +6,13 @@ os:
 - linux
 dist:
 # take the latest supported LTS version
-- trusty
+- xenial
```
elbaraim: why not Ubuntu 18.04 ? that is the latest LTS available

.travis.yml
```diff
@@ -6,13 +6,13 @@ os:
...
 
 # install dependencies
 
 before_install:
 # install latest r
-- sudo sh -c 'echo "deb http://cran.rstudio.com/bin/linux/ubuntu trusty/" >> /etc/apt/sources.list'
+- sudo sh -c 'echo "deb https://cloud.r-project.org/bin/linux/ubuntu xenial/" >> /etc/apt/sources.list'
```
elbaraim: same here (?)
you could use: https://cloud.r-project.org/bin/linux/ubuntu  bionic-cran35

pyabc/external.py
```diff
@@ -9,6 +9,13 @@
     The rpy2 package needs to be installed to interface with the R language.
     Installation of rpy2 is optional if R support is not required.
     See also :ref:`installation of optional dependencies <install-optional>`.
+
+.. note::
+
+    Support of R via rpy2 is considered experimental. Currently, an rpy2
+    version <= 2.9.5 is required. Versions >= 3.0.0 do not work yet properly
```
elbaraim: add this version checking already in setup.py ?
Maybe this is already there

.travis.yml
yannikschaelte: on travis, only 12/14/16.04 are available (it's always a bit behind).
I would actually like to also test on trusty still (so I'll keep this in mind, just doesn't work with R), but since standard support for it has stopped last month anyway, it's maybe no problem to move on.

pyabc/external.py
yannikschaelte: yes, in setup.py we currently require rpy2<=2.9.5 already. However, the user might have updated it somewhere else at some point, so the info here might be useful.

PR #119, opened 2019-06-06
Improve confidence interval plots

yannikschaelte:
* Minor modifications of the confidence_interval plots
* Add a new plot that can take multiple histories, and per default plots intervals for the last iteration for all into one plot


pyabc/visualization/confidence.py
```diff
@@ -38,11 +63,11 @@ def plot_confidence_intervals(
         # extract all parameter names
         df, _ = history.get_distribution(m=m)
         par_names = list(df.columns.values)
-
-    # dimensions
```
EmadAlamoudi: why deleting this comment

PR #121, opened 2019-06-09
Confidence intervals -> credible intervals + confidences -> levels

### LINKED ISSUE #100: Confidence intervals -> credible intervals
In visualization: Called it confidence intervals now, but in a Bayesian context should better call it credible intervals.

EmadAlamoudi:
Closes #100


pyabc/visualization/confidence.py
```diff
@@ -118,7 +118,7 @@ def plot_confidence_intervals(
                 kde_max_1d[i_par, i_t] = _kde_max_1d_pnt[par]
             # confidences
             for i_c, confidence in enumerate(confidences):
```
yannikschaelte: Maybe we can rename the `confidences` argument to something better? Maybe `levels`?

EmadAlamoudi: Done.

Issue #94, opened 2019-03-29
Implement a general interface to external simulators (via csv file exchange?)


yannikschaelte:
A way to call arbitrary external simulators (in R, matlab, java, ...) might be as follows: In the model simulate function, a command line script is executed, which reads parameters from input or from a csv file, and returns the simulations sum stats as a csv file on the file system. Then, this csv can be read in via pandas in pyabc.

Things to take care of:
* Error returns: Catch error output if the external command fails. This is not trivial as it runs in a different process.
* Generate temporary ids for the files generated to be unique.
* Delete files after usage.
* Where to save things in order not to mess up dirs when not everything gets deleted? Probably tempfile.gettempdir()
* Time overhead due to way via file system, which can be slow on some distributed systems.

This will enable pyabc to work reliably with other languages and circumvent the use of error-prone libraries (maybe also as a replacement of #66). There might be certain limitations as to what can be saved as a csv, but it's worth a try.

yannikschaelte:
closed via #130 . should work for some simple simulators. if there are other requirements (like: adding additional arguments or another argument style to the script, or directly reading in summary statistics), this would have to be on a by-need base.

PR #125, opened 2019-06-24
Feature external simulators

yannikschaelte:
This adressed #94. It offers a basic implementation of calling external simulation etc. routines, writing things to file, and finally reading them in at some point. Further details like csv exchange still need to be figured out as required in applications.


doc/examples/external_simulators.ipynb
```diff
@@ -0,0 +1,238 @@
...
+    "# Script based external simulators"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In this notebook, we demonstrate the usage of the script based external simulators, summary statistics, and distance functions features.\n",
+    "\n",
+    "These allow to use near-arbitrary programing languages and output for pyABC. The main concept is that all communication happens via the file system. This comes at the cost of a considerable overhead, making this feature not optimal for models with a low simulation time. For more expensive models, the overhead should be negligible."
```
elbaraim: "This comes at the cost of a considerable overhead, making this feature not optimal for models with a low simulation time. For more expensive models, the overhead should be negligible"

I think it is the other way around, right?
low simulation time -> good
expensive models -> considerable overhead

doc/examples/external_simulators.ipynb
```diff
@@ -0,0 +1,238 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This notebooke replicates the using_R notebook, but circumvents usage of the rpy2 package. Note the considerably longer runtime though."
```
elbaraim: typo: "notebook"

elbaraim: I would say instead of "Note the considerably longer runtime though." -> "For R scripts we encourage users to use r2py" (or something like this)

doc/examples/external_simulators.ipynb
```diff
@@ -0,0 +1,238 @@
...
+   "source": [
+    "import pyabc\n",
+    "import pyabc.external"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Here, we define model, summary statistics and distance. Note that if possible, alternatively FileIdSumStat can be used to read in the summary statistics directly to python and then use a python based distance function."
```
elbaraim: maybe you can include this piece of code calling "FileIdSumStat" as commented block-code to illustrate how. Or is there any other example that already covers this feature?

doc/examples/external_simulators.ipynb
```diff
@@ -0,0 +1,238 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "model = pyabc.external.ExternalModel(executable=\"Rscript\", file=\"rmodel/model.r\")\n",
+    "sumstat = pyabc.external.ExternalSumStat(executable=\"Rscript\", file=\"rmodel/sumstat.r\")\n",
+    "distance = pyabc.external.ExternalDistance(executable=\"Rscript\", file=\"rmodel/distance.r\")\n",
```
elbaraim: can you add here a little comment indicating that the experimental data is already included in the definition of "distance.r" ? Otherwise the user needs to check the using_R script (which I did, because I did not know where the data was coming from in the last plot). May be good to shortly indicate this.

doc/examples/rmodel/distance.r
```diff
@@ -0,0 +1,28 @@
...
+    pars = c(pars, s[[1]][2])
+}
+
+names(pars) = names_
+
+
+distance <- function(pars){
+    sumstat_0 = readRDS(pars$sumstat_0)
+    # insert observed data here directly
+    # note: by design, pyabc can currently not completely make sure
```
elbaraim: does this mean this approach may not work as it should?

test/test_external.py
```diff
@@ -1,18 +1,16 @@
...
 import pyabc
-
-
-r_file = "doc/examples/myRModel.R"
+import pyabc.external
 
 
 def test_r():
     """
     This is basically just the using_R notebook.
```
elbaraim: I would rather say "this is reproducing the using_R notebook" (?) because is not exactly the same from a coding point of view.

test/test_external.py
```diff
@@ -1,18 +1,16 @@
...
 
-from pyabc.external import R
 import pyabc
-
-
-r_file = "doc/examples/myRModel.R"
+import pyabc.external
 
 
 def test_r():
```
elbaraim: may be good to include additional languages at some point, e.g., Morpheus (once it is working)

yannikschaelte: yes! for morpheus, we will even have a complete extra file (currently in morpheus_interface branch), but it should kind of extend the general ExternalModel class. then also tests need to be added.

elbaraim: some model in C ? or matlab? :no_good_man:

yannikschaelte: matlab we probably can't install on travis, in C I don't have a model. You have one in Java, right? You might very much want to add that here :). I think you mentioned that we'll need to pass additional arguments then. Can you name a typical command line call that would be used? Then we can maybe add functionality to cover such setups.

doc/examples/external_simulators.ipynb
yannikschaelte: no, it's the way I meant it: if the simulation time is low, then the relative impact of always going via the file system is high, whereas you will hardly notice the overhead when the model simulation already takes minutes.

elbaraim: ah! okay I see what you mean :)

doc/examples/external_simulators.ipynb
yannikschaelte: there is no example covering this feature. I removed the FilIdSumStat actually in the meantime, because it did not make too much sense (the functionality of reading in data should already be in the model, but that's rather application-specific).

test/test_external.py
elbaraim: Yes, i was thinking about this example.
normally I call scala via:
`call(["/usr/bin/taskset","-c","0",(pathMwn+"/mvnw"), "scala:run", "\"-DaddArgs={}\"".format(params_argument)])`

doc/examples/rmodel/distance.r
yannikschaelte: currently, we always use the first argument for the simulated, and the second for the measured data. thus, it works at the moment. however, the structure of these arguments makes it not really clear. I think we should rename the arguments to ``simulated_sum_stat`` and ``measured_sum_stat`` at some point to ensure clarity.

btw: if the user mixes the order or sth changes, that will be immediately noticed by a huge error, so no problem.

yannikschaelte: see #128

test/test_external.py
yannikschaelte: added the possibility to define custom arguments in addition now.

yannikschaelte:
> Overall it looks good!
> However I have the following question:
> 
> * Using some external simulators, e.g. based on Java, by default would take all threads in the memory available. Do you plan to include an additional input configuration for this? (e.g. taskset)
> * Make sure that when installing pyABC from pip (or from git) the "test/run_notebooks.sh" file is an executable and not a plain text file, otherwise it will lead to errors.
> 
> The rest are minor comments :)

Thanks! The first point should be addressed now (need to test on java though yet), for the second: Yes, it has to be executable by rights. I hope that git takes care of that.

Issue #101, opened 2019-04-10
Easy extraction of reference values from db


yannikschaelte:
In light of #99, there should be convenience functions to extract the reference values from the history, as exists already for the observed summary statistics. gt_model and gt_par (ground truth model and parameters) should be straightforward.

PR #126, opened 2019-06-24
Extract reference value

yannikschaelte:
addresses #101

### DIFF
--- a/pyabc/storage/history.py
+++ b/pyabc/storage/history.py
@@ -1132,9 +1132,10 @@ class History:
 
 
     @with_session
-    def get_parameter_reference_value(self,refval:str) -> dict:
+    def get_parameter_reference_value(self, refval: str) -> dict:
         """
-        Create a pyabc.Parameter object containing the parameter that is equal to the reference value
+        Create a pyabc.Parameter object containing the parameter that
+        is equal to the reference value
 
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


pyabc/storage/history.py
```diff
@@ -1129,3 +1129,30 @@ def get_population_extended(self, *, m: Union[int, None] = None,
...
+        """
+        parameter = (self._session.query(Parameter)
+                     .join(Particle)
+                     .join(Model)
+                     .join(Population)
+                     .join(ABCSMC)
+                     .filter(ABCSMC.id == self.id)
+                     .filter(Parameter.name == refval)
+                     .all())
+        return {parameter[0].name: parameter[0].value}
```
yannikschaelte: I think this is not correct yet when we have more than 1 parameter.

pyabc/storage/history.py
```diff
@@ -1129,3 +1129,30 @@ def get_population_extended(self, *, m: Union[int, None] = None,
                 df = df_tidy
 
         return df
+
+    @with_session
+    def get_parameter_reference_value(self, refval: str) -> dict:
```
yannikschaelte: I would rather recommend a function that does not take `refval` as an argument, but only self, and then returns the full parameter, e.g. `{'meanX': 3.5, 'meanY': 42,, 'z': 23.513}`

pyabc/visualization/histogram.py
```diff
@@ -60,6 +62,8 @@ def plot_histogram_1d_lowlevel(
         range_ = (xmin, xmax)
     else:
         range_ = None
+    if refval is not None:
+        ax.axvline(refval[x], color='C1', linestyle='dashed')
```
elbaraim: possibility to change refval color? why not in black?

yannikschaelte: not black, because we also use it on the (black) scatter plots. But I agree that 'C1' is not ideal. Until now, it always worked, so I would leave it until we encounter an application problem.

PR #133, opened 2019-07-28
Redis catch errors

EmadAlamoudi:
Also,
* MinMaxDistanceFunction -> MinMaxDistance in test_abc_smc_algorithm & test_samplerperf
* PercentileDistanceFunction -> PercentileDistance in test_abc_smc_algorithm & test_samplerperf


EmadAlamoudi:
@yannikschaelte, can you please look at the Travis failure. I think it is something related to R & external module.

test/test_samplers.py
```diff
@@ -225,3 +227,23 @@ def simulate_one():
...
+    def model(pars):
+        print(f"Model for {pars}")
+        if np.random.uniform() < 0.1:
+            raise ValueError("error")
+        return {'s0': pars['p0'] + 0.2 * np.random.uniform()}
+
+    def distance(s0, s1):
+        return abs(s0['s0'] - s1['s0'])
+    prior = Distribution(p0=RV("uniform", 0, 10))
+    sampler = RedisEvalParallelSampler(port=8775)
```
yannikschaelte: When you have the RedisEvalParallelSampler here, no workers are submitted. Instead, you should use the RedisEvalParallelSamplerServerStarter here.

test/test_samplers.py
```diff
@@ -225,3 +227,23 @@ def simulate_one():
...
+    abc = ABCSMC(model, prior, distance, sampler=sampler)
+    db_file = "sqlite:///" + os.path.join(tempfile.gettempdir(), "test.db")
+    data = {'s0': 2.8}
+    abc.new(db_file, data)
+    with pytest.raises(Exception):
```
yannikschaelte: I would assume that in the new implementation nothing gets raised, i.e. you can just remove the ``with pytest.raises`` and just call run

test/test_samplers.py
```diff
@@ -225,3 +227,23 @@ def simulate_one():
     sample = sampler.sample_until_n_accepted(10, simulate_one)
     assert 10 == len(sample.get_accepted_population())
     sampler.cleanup()
+
+
+def test_redis_catch_error():
+
+    def model(pars):
+        print(f"Model for {pars}")
```
yannikschaelte: No ``print`` needed here.

yannikschaelte:
> @yannikschaelte, can you please look at the Travis failure. I think it is something related to R & external module.

Oh yes. There was an update of numpy (1.16.2 worked, 1.17.0 not) which cannot handle some data type any more, it seems. Again Rpy2 seems to fail. This is really tiresome.

test/test_samplers.py
EmadAlamoudi: When removing `with pytest.raises`, the test method will fail then since the model create error

yannikschaelte: But the idea was that any errors are not propagated, but caught in the ``try-catch``, right?

pyabc/sampler/redis_eps/cli.py
```diff
@@ -120,7 +120,11 @@ def work_on_population(redis: StrictRedis,
         # make batch_size attempts
         for n_batched in range(batch_size):
             # simulate
-            new_sim = simulate_one()
+            try:
+                new_sim = simulate_one()
+            except Exception as e:
+                logger.warning(f"Redis worker number:{n_worker} fail."
```
yannikschaelte:
```suggestion
                logger.warning(f"Redis worker number {n_worker} failed."
```

PR #140, opened 2019-08-07
Maintenance stuff, in particular rpy2 3.1.0

yannikschaelte:
- fix R installation by giving full key
- port to rpy2 version >= 3.0.0, which involved some API changes.
- currently, install the very latest rpy2 version from bitbucket, until 3.1.0 gets released (current 3.0.5 does not support multithreading)
- add full license info to setup.py


pyabc/external/r_rpy2.py
```diff
@@ -11,20 +11,21 @@
...
-    Support of R via rpy2 is considered experimental. Currently, an rpy2
-    version <= 2.9.5 is required. Versions >= 3.0.0 do not work yet properly
-    for various reasons (see #116). R versions 3.4 and 3.5 have shown to work.
+    Support of R via rpy2 is considered experimental for various reasons
+    (see #116).
     Should this not work on your system, consider accessing R script-based.
 """
 
 try:
+    import rpy2.robjects
```
EmadAlamoudi: why two imports for the same library?

yannikschaelte: The second one is just as a convenience in order ton not always have to write ``rpy2.robjects.r`` but just ``r``. But I will simplify that now, actually right.

PR #152, opened 2019-08-13
Feature noisyacceptor

yannikschaelte:
Introduction of an acceptor that can directly account for measurement noise in an efficient manner.


yannikschaelte:
discontinued here since the branch got outdated.

PR #165, opened 2019-09-16
Data plot

yannikschaelte:
Let's bring this to completion soon. @EmadAlamoudi could you 
* finalize your code,
* add a section in some jupyter notebook on this, showing how it looks in practice,
* provide some tests?


EmadAlamoudi:
I'm not sure if the version that I have for the cli file is old. Need to be checked.

EmadAlamoudi:
There is an error raised in the test_transition function. Not sure what is the cause.

yannikschaelte:
> There is an error raised in the test_transition function. Not sure what is the cause.

yeah that test sometimes fails due to stochastic effects. I restarted it now and created an issue, #167

pyabc/visualization/data_plot.py
```diff
@@ -0,0 +1,70 @@
+import pandas as pd
+import numpy as np
+import matplotlib.pyplot as plt
+
+
+def plot_data(obs_data: dict, sim_data: dict, key=None):
+    fig = plt.figure()
```
yannikschaelte: please add a docstring describing the functionality

pyabc/visualization/data_plot.py
```diff
@@ -0,0 +1,70 @@
...
+        plot_col_size = 2
+    elif len(obs_data) <= 9:
+        plot_row_size = 3
+        plot_col_size = 3
+    elif len(obs_data) <= 16:
+        plot_row_size = 4
+        plot_col_size = 4
+    else:
+        print("Data length should be equal or less than 16 ")
+        return -1
```
yannikschaelte: to my knowledge, return codes are uncommon in python. I would just raise the warning and then ``return``.
Please also use logger.warn (see how to use a logger here: https://github.com/ICB-DCM/pyABC/blob/master/pyabc/distance/distance.py#L12) instead of all print commands.

pyabc/visualization/data_plot.py
```diff
@@ -0,0 +1,70 @@
...
+            print('The selected data type is '
+                  'not yet supported. Try to use '
+                  'Pandas.Dataframe, 1d numpy.array, '
+                  'or 2d numpy.array')
+            return -1
+        plt.xlabel('Time $t$')
+        plt.ylabel('Measurement $Y$')
+        plt.title(obs_key)
+        plt.legend()
+        fig.suptitle("Observed vs. simulated data", fontsize=16)
```
yannikschaelte: probably, you should return the ``plt`` instance here (background: then the user can post-modify the plots), see also the other plotting routines.

test/visualization/test_visualization.py
```diff
@@ -87,3 +87,13 @@ def test_credible_intervals():
 
 def test_model_probabilities():
     pyabc.visualization.plot_model_probabilities(histories[0])
+
+
+def test_data_plot():
```
yannikschaelte: this would be enough tests for me at the moment (in particular, since my own tests of visualization routines also just execute code ^^)

doc/examples/data_plots.ipynb
```diff
@@ -0,0 +1,144 @@
...
+   "metadata": {},
+   "source": [
+    "## Define summary statistics for observed data"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "First, we will define some arbitrary summary statistics to be used in this example. We will define four different summary statistics with different data types. Data types of the value of the summary statistic can be 1d numpy array, 2d numpy array, dictionary."
```
elbaraim: The third example is not really a dictionary, right? As it is converted to a pandas dataframe (?)
- change in the explanation text

doc/examples/data_plots.ipynb
```diff
@@ -0,0 +1,144 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "observed = {'data 1': np.array([1,1,2,2,3,3,4,4]),\n",
+    "            'data 2': pd.DataFrame([{'measurement': [1,2,3,4,1,2,3,4]}]),\n",
+    "            'data 3': np.array([[4,4,3,3,2,2,1,1],[4,4,3,3,2,2,1,1]]),\n",
```
elbaraim: in "data 3" is supposed to be 2 types of summary statistics?

pyabc/visualization/data.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+    Plot summary statistic data.
+
+    Parameters
+    ----------
+
+    obs_data: dict
+        A dictionary for the summary statistic of the observed data,
+        where keys represent the summary statistic name and values represent
+        the data itself. The values can be represented as pandas dataframe,
```
elbaraim: here in the function help says "represented as pandas dataframe" not dictionary. Please check consistency

pyabc/visualization/data.py
```diff
@@ -0,0 +1,103 @@
...
+        1d numpy array, or 2d numpy array.
+    sim_data: dict
+        A dictionary for the summary statistic of the simulated data,
+        where keys represent the summary statistic name and values represent
+        the data itself. The values can be represented as pandas dataframe,
+        1d numpy array, or 2d numpy array.
+    key: Union[str, int], optional
+        A specified specific summary statistic data, if None,
```
elbaraim: "specified specific" ?? probably this is some typo/mistake ?

pyabc/visualization/data_plot.py
yannikschaelte: sorry, I meant the axes object. changed that now already.

doc/examples/data_plots.ipynb
yannikschaelte: yeah no, just some arbitrary 2d data. I will make these more versatile.

pyabc/visualization/data.py
yannikschaelte: yes, it should always be pandas dataframe, entries cannot be dicts.

pyabc/visualization/data.py
```diff
@@ -0,0 +1,103 @@
...
+            ax.plot(sim_value, '-x', color="C0", label='Simulation')
+            ax.plot(obs_value, '-x', color="C1", label='Data')
+        else:
+            logger.info("The selected data type is "
+                        "not yet supported. Try to use "
+                        "Pandas.Dataframe, 1d numpy.array, "
+                        "or 2d numpy.array. Found = {}.".format(type(obs)))
+
+        # finalize axes
+        ax.set_title(obs_key)
```
elbaraim: what about the axis labels? I think it is better to show on the axis labels the name of the summary statistic instead as title.
- in the `ax.plot()` lines what is supposed to be the X axis there? Time in arbitrary units? Number of data points?
- in the `ax.plot()` lines the Y axis can be named as `obs_key` instead and remove the title

elbaraim: regarding the plotting of 2D numpy array data:
wouldn't be more intuitive to plot the 1st dimension of observed data vs  1st dimension simulated and then 2nd vs 2nd ?
otherwise i do not get the visual comparison... (?)

test/visualization/test_visualization.py
```diff
@@ -87,3 +87,13 @@ def test_credible_intervals():
...
+    obs_dict = {1: 0.7}
+    sim_dict = {1: 6.5}
+    data_plot.plot_data(obs_dict, sim_dict)
+    for i in range(18):
```
elbaraim: for the sake of the test a smaller `range` is also valid :wink:

pyabc/visualization/data.py
```diff
@@ -0,0 +1,103 @@
...
+    -------
+
+    ax: Axis of the generated plot.
+    """
+    # check if user specified a specific key to be printed
+    if key is not None:
+        obs_data = {key: obs_data[key]}
+        sim_data = {key: sim_data[key]}
+
+    # get number of rows and columns
```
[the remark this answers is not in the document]
yannikschaelte: that's a good idea. worked it in now, seems to work just fine.

pyabc/visualization/data.py
yannikschaelte: the problem about the x axes is that we usually have not information about what is the label. can be time, can be anything. Maybe we just use some generic "Index" as label for the moment. One could of course allow the user to pass labels to the plot function, but I don't want to do that at the moment.

elbaraim: a generic index would be fine, yes, better than no labels :)

elbaraim: what about my comment regarding the 2D data?

yannikschaelte: good point. not sure how to best plot 2d data. eventually probably application will tell, but I think I will go with your approach.

yannikschaelte:
still not super convinced how applicable the plots will be, in particular 2d, but i would leave it as it is now, a bit rudimentary though. @elbaraim wanna have another look?

pyabc/visualization/data.py
```diff
@@ -0,0 +1,111 @@
...
+            ax.set_xlabel("Data")
+            ax.set_ylabel("Simulation")
+        else:
+            logger.info(f"Data type {type(obs)} for key {obs_key} is "
+                        f"not supported.")
+            # remove not needed axis
+            ax.axis('off')
+
+        # finalize axes
+        ax.set_title(str(obs_key))
```
elbaraim: what about having only one title for all the subplots on the top called "Summary Statistics" then you avoid repetition of axis label naming and subplot title?

doc/examples/data_plots.ipynb
```diff
@@ -0,0 +1,150 @@
...
+      ]
+     },
+     "metadata": {
+      "needs_background": "light"
+     },
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "plot_data(observed, simulated)\n",
```
elbaraim: increase `figsize` for the example visualization ? :wink:

pyabc/visualization/data.py
yannikschaelte: some subplots don't have the key as axis label, so it might be better to have it as subplot label always.

PR #173, opened 2019-09-28
Aggregated distances and user-defined scaling factors

yannikschaelte:
* Introduce new `AggregatedDistance` and `AdaptiveAggregatedDistance` classes to combine different distance metrics into one overall metric.
* Define factors additional to the weights that allow to handle the case when one data type is larger in number of summary statistic values.


doc/examples/aggregated_distances.ipynb
```diff
@@ -0,0 +1,654 @@
...
+     "metadata": {
+      "needs_background": "light"
+     },
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "for t in range(h.max_t, h.max_t+1):\n",
+    "    ax = pyabc.visualization.plot_kde_2d(*h.get_distribution(m=0, t=t),\n",
+    "                                         \"theta1\", \"theta2\",\n",
```
elbaraim: The plot legend is "$\\Theta$" therefore I would also plot as X and Y axis labels "$\\Theta_1$" and "$\\Theta_2$"
or change the legend name to "theta" (consistency)

doc/examples/aggregated_distances.ipynb
```diff
@@ -0,0 +1,654 @@
...
+      "{'theta1': 0.574942476350519, 'theta2': 0.5605028742014329} 2.4344988300943373 0.24\n",
+      "{'theta1': 0.0820849986238988, 'theta2': 0.1353352832366127} 0.0\n"
+     ]
+    }
+   ],
+   "source": [
+    "for _ in range(10):\n",
+    "    p = dict(parameter_prior.rvs())\n",
+    "    sim = model(p)\n",
+    "    print(p, cr_l1_distance(sim, observation), gauss_ks_distance(sim, observation))\n",
```
elbaraim: - can you instead generate a plot? it would be easier for interpretation
- or a pandas dataframe with column names?
if not, then I would include and additional printed line before the last print statement as follows:
`print("True parameters:")`

pyabc/distance/scales.py
```diff
@@ -19,113 +19,134 @@
...
+
+
+def span(data, **kwargs):
+    """
+    Compute the difference of largest and smallest data point.
+    """
+    return max(data) - min(data)
+
+
+def mean(data, **kwargs):
```
elbaraim: Why do you define this function if in the previous instances you use `np.mean` instead of this new defined `mean` function?

pyabc/distance/scales.py
```diff
@@ -19,113 +19,134 @@
...
 
 
-def bias(**kwargs):
+def bias(data, x_0, **kwargs):
     """
     Bias of sample to observed value.
     """
-    data = np.asarray(kwargs['data'])
-    x_0 = kwargs['x_0']
     bias = np.abs(np.mean(data) - x_0)
```
elbaraim: here you could use the `mean` function defined down below (and following)

pyabc/distance/scales.py
```diff
@@ -19,113 +19,134 @@
...
 
-def median_absolute_deviation_to_observation(**kwargs):
+def median_absolute_deviation_to_observation(data, x_0, **kwargs):
     """
     Median absolute deviation of data w.r.t. the observation x_0.
     """
-    data = np.asarray(kwargs['data'])
-    x_0 = kwargs['x_0']
+    data = np.array(data)
     mado = np.median(np.abs(data - x_0))
```
elbaraim: here you can use the `median` (same as for the `mean`)
- check all the rest

pyabc/distance/scales.py
```diff
@@ -19,113 +19,134 @@
...
-def standard_deviation_to_observation(**kwargs):
+def standard_deviation_to_observation(data, x_0, **kwargs):
     """
     Standard deviation of absolute deviations of the data w.r.t.
     the observation x_0.
     """
-    data = np.asarray(kwargs['data'])
-    x_0 = kwargs['x_0']
+    data = np.array(data)
     stdo = np.std(np.abs(data - x_0))
```
elbaraim: here you can use `standard_deviation` instead

pyabc/__init__.py
```diff
@@ -26,6 +26,8 @@
     SimpleFunctionDistance,
     PNormDistance,
     AdaptivePNormDistance,
+    AggregatedDistance,
+    AdaptiveAggregatedDistance,
```
EmadAlamoudi: I don't know, but can't we combine both (AggregatedDistance and AdaptiveAggregatedDistance) functions in one and then use the adaptive flag to decide?

pyabc/distance/scales.py
```diff
@@ -19,113 +19,134 @@
...
 
 
-def standard_deviation(**kwargs):
+def standard_deviation(data, **kwargs):
     """
     Calculate the sample `standard deviation (SD)
     <https://en.wikipedia.org/wiki/Standard_deviation/>`_.
     """
-    data = np.asarray(kwargs['data'])
     std = np.std(data)
```
elbaraim: Regarding std calculation: by default in `np.std` the argument `ddof` is set to zero. Don't we prefer to have unbiased estimation of the std?
Details in the notes section https://docs.scipy.org/doc/numpy/reference/generated/numpy.std.html

pyabc/visualization/sample.py
```diff
@@ -83,3 +83,89 @@ def plot_sample_numbers(
...
+        # be included here to be fair against methods not requiring
+        # calibration
+        samples.append(np.sum(history.get_all_populations()['samples']))
+    samples = np.array(samples)
+
+    # apply scale
+    ylabel = "Total samples"
+    if yscale == 'log':
+        samples = np.log(samples)
+        ylabel = "log(" + ylabel + ")"
```
elbaraim: is it normally used? plotting axes in natural log?

pyabc/distance/scales.py
```diff
@@ -19,113 +19,134 @@
...
     return std
 
 
-def bias(**kwargs):
+def bias(data, x_0, **kwargs):
     """
     Bias of sample to observed value.
```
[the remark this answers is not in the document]
yannikschaelte: right. I added a description now in the top of the file, because otherwise I would need to repeat the same text again and again. This could be solved by subclassing, but for the moment I hope it will be fine.

pyabc/__init__.py
yannikschaelte: Good point. Both have their advantages. I wanted to separate both to have clearly all basic functionality for weighting etc. in the base class, and all "adapting" functionality in the subclass, like for (Adaptive)PNormDistance. Note that ``adaptive`` means whether to only pre-calibrate the weights, or adapt them after each iteration. The name of the class might be a bit misleading, but I hope that's fine.

yannikschaelte:
> Some changes to be reviewed.
> 
> * Highlighted: why is needed the definition of `mean`, `median` and `standard_deviation` functions?

What do you mean here? We define shallow wrappers of mean, median etc. around the respective numpy functions, in order to have a common interface which we can use from inside the Distance functions.

elbaraim:
> > Some changes to be reviewed.
> > 
> > * Highlighted: why is needed the definition of `mean`, `median` and `standard_deviation` functions?
> 
> What do you mean here? We define shallow wrappers of mean, median etc. around the respective numpy functions, in order to have a common interface which we can use from inside the Distance functions.

okay, but then shouldn't be these used inside the other quantities, intead of e.g. `np.mean` (see more details in the review)

doc/examples/adaptive_distances.ipynb
```diff
@@ -92,17 +100,17 @@
...
-      "INFO:ABC:t:0 eps:8.323532755675517\n",
-      "INFO:ABC:t:1 eps:4.91440739009507\n",
-      "INFO:ABC:t:2 eps:3.4217303156404038\n",
-      "INFO:ABC:t:3 eps:2.370056613878034\n",
-      "INFO:ABC:t:4 eps:1.6268813652396334\n",
-      "INFO:ABC:t:5 eps:1.186727799009716\n",
-      "INFO:ABC:t:6 eps:0.8548463991669453\n",
-      "INFO:ABC:t:7 eps:0.5934272512057183\n",
-      "INFO:History:Done <ABCSMC(id=3, start_time=2019-05-01 16:05:34.676351, end_time=2019-05-01 16:07:50.875925)>\n"
+      "INFO:History:Start <ABCSMC(id=53, start_time=2019-09-28 20:24:52.489329, end_time=None)>\n",
```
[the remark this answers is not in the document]
yannikschaelte: yeah to be honest I don't care :laughing: also depends on in which order the notebooks are run.

doc/examples/adaptive_distances.ipynb
```diff
@@ -154,14 +152,18 @@
...
     "\n",
-    "fig, ax = pyplot.subplots()\n",
-    "for t in range(history0.max_t + 1):\n",
-    "    df, w = history0.get_distribution(m=0, t=t)\n",
-    "    pyabc.visualization.plot_kde_1d(df, w, xmin=0, xmax=10,\n",
-    "                                   x='theta', ax=ax,\n",
-    "                                   label=\"PDF t={}\".format(t))\n",
-    "ax.axvline(theta_true, color='k', linestyle='dashed', label=\"True value\")\n",
-    "ax.legend()"
+    "def plot(history):\n",
```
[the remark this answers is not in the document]
yannikschaelte: good point. will do.

doc/examples/adaptive_distances.ipynb
```diff
@@ -253,43 +284,256 @@
...
     "\n",
-    "fig, ax = pyplot.subplots()\n",
-    "for t in range(history1.max_t + 1):\n",
-    "    df, w = history1.get_distribution(m=0, t=t)\n",
-    "    pyabc.visualization.plot_kde_1d(df, w, xmin=0, xmax=10,\n",
-    "                                   x='theta', ax=ax,\n",
-    "                                   label=\"PDF t={}\".format(t))\n",
-    "ax.axvline(theta_true, color='k', linestyle='dashed', label=\"True value\")\n",
-    "ax.legend()"
+    "def plot(history):\n",
```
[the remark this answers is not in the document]
yannikschaelte: this was, because the three subchapters define kind of independent things (still under the adaptive distances context, thus in the same notebook). so I wanted to have everything the user needs to run a subpart in the chapter.

doc/examples/adaptive_distances.ipynb
```diff
@@ -301,15 +545,430 @@
...
...nGYLkqovxj0wSpEHko2TJ09SdeQIE2vrUAonCOF6RRXimyRJSBqNK2ldS4VG61rq9SiDg1GFBKMMalkGB6MMDESSjSa7HHcaPUYBhTd8LgIe/rY8Qgi7JEn1QAgwABCSJO0GwoB1Qoj/c+sJJEl6AXgBIDY29q7B1G29hrWk6a552osm0pfAeW0zSvS0Bf3ly5cJDAxk0aJF5ObmMm3aNN566622u7sKAeXnXeKRvQ3Kz7m29xoCE34M/aZAxBDS049Qf/Qozz65AmVnWKb4BEH0SCz2BKq3XkSY7QQ93h/fwW9C8XxXiajwazj/GZxa7donpD/0nwH9p7vERdV1c6XcDadTcCy3mm2Zpew+X0Z1kxWdWsGUQeHMHtqbyQPD8dV6ZlYIZ3MztrIy7OXl2ErLsFdV4qirw1Fb51peT7W1OBob4YaBiKG4qjAm3uG4Alf9tUOtQanzRT/6FRT6IQzN/iuqMC0n4+JoNvgyqaAQ3Y1taBJISAgEwmpDWK2uZLG0rjubm12/y1uRJFQREWj79UPbrx+ahH6t60rZHt9teKvRowoYD6QAJmBfy5D/fTdmEkK8D7wPLjsVN8XSIbzFgt5ut3Po0CHOnDlDbGwsjz/+OKtXr753PFYTnNsAX/8PVGQBkqsEMOM3MGguBPdpzVpWVsZXX33FiBEjOs17SwiB8UgJ9TtyUQVpCX4uGU3vFuv4vpNcCVyN+ZUXIfcgXPkCTnwAx94FjcGVp/906D8T/Ht3Slztif86qW99SVmDGR+1kimJ4cwZ2ptJA8PQa7pGQBz19Viu5WC5dhXrtRyseXkuASkrw1FXd1t+SatFGRiIMigIZWAg2gEDUAYGoPTzR+Hr+00ytCz1elfSapF8fKi2wt7cenZdquGr3BocTsH4CH/erFWgS/shs/95BLFXs9m8eTM7Hu7N4sWL29WGJxwOl8DV1GCvrsFRW4O9pgZHdQ3WokKsV69Re+oUorm5dR9lWCg+w4fjm5qKITUVdVyc15X+uivuNHosBmJu+Bzdsu1OeYpa2kkCcL3IFAEHhRBVAJIk7QAeAvZxn7S1BNHZeIsFfXR0NMnJya39/BcuXMixY8e+XUzqClwP5FNrwFwHvYbCnP8LifPAEH5bdqfTybZt2/Dx8WmtSusoTouD2k1XaD5biS4xmODHBn771LoKBfQa7EpjXnK1u+Qe
```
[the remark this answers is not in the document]
yannikschaelte: made that before already, will clarify

pyabc/distance/scales.py
yannikschaelte: yeah, one could do that, but since those functions are just shallow wrappers around the numpy functionality, I don't really see the reason (and minimally increases computation time). the problem those wrapper functions solve is that some of the scale functions e.g. requre x_0, some only data, so I defined a common interface by which all functions can be called, with additional input values just being ignored.

yannikschaelte: also, calling the own function adds another layer of uncertainty, because one has to make sure that the code in those functions does not change in some way. numpy in contrast will definitely not change.

pyabc/visualization/sample.py
yannikschaelte: don't know, i never did :smile: but doesn't hurt to allow it.

doc/examples/aggregated_distances.ipynb
```diff
@@ -0,0 +1,503 @@
...
...lpkXfgcSfiyIu6jRg6NdIS8ihIE9rcjmTW02mh38PptfcRWGgHwlz5qBLN30eRq9G3nwzthVL00N4y2aaIhx3CYpo3M9EfG94MvbvYHB5XYxb85tJyUvh2b+eRSu1LOq+CG977xv79Hodp3eHEdCsJTlLlqHLzMTrtVkl9j1odXo2Ho9n4Jd7GPntfiKvZjKtVxDhMx5iVt+G+DjbotfpiT2Thl9Dt+rXh3EX07hLd2zsHdi7ZjlSr8crwBkkXLto+g1bJVTM6zSPpl7Nmd0lCW16OonvvW+WHV2DPPnhidaszg7hVfUU5NXD8PNQKMgqO7NClaCIxv3KP5/CH5MMbkFGrQbrsoMQ5WpymbB9Atdyr/FFty+o61L3P/svHT9KdloqDwQ2JG3FClyHP4ZNgwa3lHMhOYf3/zxDh/d2MGH5YTLzNMwb3IR/XgllQmj9/8zkTryYRWGe1ixX6AplY2ljQ4dho7h0/Aj71q3Cs44TCEgsZZJfcdhY2LCg2wI09WqxsaMVGb/+Sta2bWaV0aFeDVaOb89WfWum8zLySgT8pAhHdeU2Ylsq3JVICTvfhb/fh0ZDYMhiUJftbqNQV8jLYS9zJvUMC7otoJlns1vSnNq5DRtHR+zW/47GwYEaL754Y19eoY5NJ+JZFRHLgQupqASEBnkyrLUf3YO9SpyUFxuZghCmuUJXMI9mvfqREHOO8NU/4xFQF1cvO7P6Na7jauPKVw99xeM5o2h6XoeYMYO6a9di5e9vchlNfJ1Z/Wx7xnynZkqejo+uLED8PMz4QHMHnE0qmIzS0rifkBK2v20QjOaj4ZFvTRIMjV7D1L8NYUBnd5hNZ9/Ot6TJz8kmOmIfdX0DyD9wAI+XX0I6ObMnOpmZ607Qeu42pqw+xrXMfKb1CmLvzIf47vHW9GrkXeos7suRqXgGOGFjf4f9SN0HCCHo/swEvOrWZ/MXH+LskU/ixcxyuTOv7VSbj3t+zseD1OTo87n84ovo88wbRlvXw4G1z3XgpEsokzQTkJf3GQZoFJY+f0ThzlKpoiGE6C2EiBJCRAshZhSz/3EhRJIQ4qhxeboy7bmvkRL+esPoePAJ6P85qMqOLaHT63jtn9cIiw1jZpuZDKo/qNh0UeG70Wk0uP9zgMKAerxDEK3e2caob/ez/sgVejb0YuX4doRN7cqE0Pp4OdmUeez8HA3
```
elbaraim: rename function to `plot_history`

elbaraim: and the rest of occurrences in this notebook

PR #179, opened 2019-10-08
Tidy up rpy2 installation on travis and readthedocs

### LINKED ISSUE #131: Add external simulators to docs
Somehow it does not appear in the API reference yet.

### LINKED ISSUE #150: Fix rpy2 import for readthedocs.io
Currently, rpy2 is not installed on readthedocs, thus compilation of the docs regarding external simulators fails.

### LINKED ISSUE #178: Re-use standard rpy2
For testing, we used a hand-crafted rpy2 from bitbucket on travis. Now that 3.2.0 has been published  on pypi, which addresses the problems we had, we should use the latest pypi one again, to be in line with what users commonly use. Should also fix #150.

yannikschaelte:
checked: passes on readthedocs

* fixes #178 (fix use standard rpy2 version)
* fixes #150 (fix readthedocs R import)
* fixes #131 (fix add external simulators to docs)


.readthedocs.yml
```diff
@@ -1,7 +1,24 @@
...
+
+# conda setup
 conda:
-    file: .environment.yml
+  environment: .rtd_conda_env.yml
 
+# python requirements
 python:
-    setup_py_install: true
+  version: 3.7
```
yannikschaelte: explanation: this was before only defined in the web interface of readthedocs, which leads to hard-to-track behavior, thus it is preferable to define everything here in the .yml file.

Issue #192, opened 2019-10-21
StochasticAcceptor: Temperature aggregation


yannikschaelte:
Currently, we only use a minimum vote, but should at least allow general functions.

yannikschaelte:
temperature aggregation imlemented via flexible functional input in pyabc.epsilon.Temperature.

Issue #61, opened 2019-02-05
StochasticAcceptor: Log scale


yannikschaelte:
Allow to have everything on a log-scale and only move to linear scale at the last point, after summing, normalizing etc.

This helps avoid numerical issues.

yannikschaelte:
Now used log as default, but also allowed for linear scale (may be preferable and less error-prone for some limited support distributions).

Issue #60, opened 2019-02-05
StochasticAcceptor: Use weighted distances


yannikschaelte:
For predicting the acceptance rate, use weighted distances instead of unweighted ones.

yannikschaelte:
Note: Even better would be the usage of using the transition weights of the next iteration. Currently not implemented yet.

yannikschaelte:
done in #197

Issue #59, opened 2019-02-05
StochasticAcceptor: Implement Sample result post-evaluation


yannikschaelte:
After a sample round has finished, check whether any assumption was hit (e.g. a distance found that is above the previously assumed maximum distance), re-evalulate the population, and possibly run some more samples.

Iterate this until sufficiently many samples accepted.

yannikschaelte:
Became obsolute with the introduction of rejection control importance sampling.

PR #197, opened 2019-10-23
Stochastic Acceptance Step

### LINKED ISSUE #57: StochasticAcceptor: Implement Stochastic Comparator

yannikschaelte:
* Introduce acceptor.StochasticAcceptor to encode the stochastic acceptance step replacing the standard uniform criterion.
* Introduce distance.StochasticKernel to encode noise distributions, with several concrete implementations already (Gauss, Binomial).
* Introduce epsilon.Temperature to capture the temperature replacing the traditional epsilons. In addition, multiple concrete TemperatureSchemes were implemented what handle the calculation of the next temperature value.

The goal of this implementation to be low-invasive, yet clear in the design and easily customizable. There are still a few flaws, 

* e.g. that StochasticKernels are not really a subclass of Distance functions concept-wise, could be generalized to a general Comparator class.
* The information/data exchange between ABCSMC, Acceptor, Distance, Epsilon is not optimal yet and should be abstracted.

Doing these things better would however require deeper changes in the infrastructure and may thus remain for later. 

Why is this no major release? I would prefer to go little steps (since this is not API breaking) and move to 0.10 series when the new functionality is mature.

Fixes #57, #58, #59, #60, #61, #192, #201


pyabc/epsilon/temperature.py
```diff
@@ -0,0 +1,502 @@
...
+    def __call__(self,
+                 t: int) -> float:
+        return self.temperatures[t]
+
+
+class TemperatureScheme:
+    """
+    A TemperatureScheme suggests the next temperature value. It is used as
+    one of potentially multiple schemes employed in the Temperature class.
+    This class is abstract.
```
elbaraim: what does this mean?

pyabc/epsilon/temperature.py
```diff
@@ -0,0 +1,502 @@
...
+        pass
+
+
+class AcceptanceRateScheme(TemperatureScheme):
+    """
+    Try to keep the acceptance rate constant at a value of
+    `target_rate`. Note that this scheme will fail to
+    reduce the temperature sufficiently in later iterations, if the
+    problem's inherent acceptance rate is lower, but it has been
+    observed to give big feasible temperature leaps in early iterations.
```
elbaraim: " it has been observed to give ... " -> can you add some examples/citation ?

pyabc/epsilon/temperature.py
```diff
@@ -0,0 +1,502 @@
...
+        # for the next proposal transition kernel.
+        weights /= np.sum(weights)
+
+        # objective function which we wish to find a root for
+        def obj(beta):
+            val = np.sum(weights * values**beta) - self.target_rate
+            return val
+
+        if obj(1) > 0:
+            # function is mon. dec., smallest possible value already > 0
```
elbaraim: is " mon. dec." an usual abbreviation in math?

pyabc/epsilon/temperature.py
```diff
@@ -0,0 +1,502 @@
...
+
+    where by default alpha=0.25.
+
+    So, in both cases the sequence of temperatures follows an exponential
+    decay, also known as a geometric progression, or a linear progression
+    in log-space.
+
+    Note that the formula is applied anew in each iteration. That has the
+    advantage that, if also other schemes are used s.t. T_{j-1} is smaller
+    than by the above, advantage can be made of this.
```
elbaraim: "That has the advantage that, ... , advantage can be made of this." This sounds funny

pyabc/epsilon/temperature.py
```diff
@@ -0,0 +1,502 @@
...
+    Requires finite `max_nr_populations`.
+
+    Note that this is similar to the `scheme_exponential_decay`, which is
+    indeed the limit for `temp_decay_exponent -> infinity`. For smaller
+    exponent, the sequence makes larger steps for low temperatures. This
+    can be useful in cases, where lower temperatures (which are usually
+    more expensive) can be traversed in few larger steps.
+    """
+
+    def __init__(self, exponent: float = 3):
```
elbaraim: why 3?

pyabc/epsilon/temperature.py
```diff
@@ -0,0 +1,502 @@
...
+
+    def __call__(self,
+                 t: int,
+                 get_weighted_distances: Callable,
+                 max_nr_populations: int,
+                 pdf_norm: float,
+                 kernel_scale: str,
+                 prev_temperature: float,
+                 acceptance_rate: float):
+        # execute functino (expensive if in calibration)
```
elbaraim: functino -> function

pyabc/epsilon/temperature.py
```diff
@@ -0,0 +1,502 @@
...
+
+        beta = beta_base + ((1. - beta_base) * 1 / t_to_go) ** 2
+
+        temperature = 1. / beta
+        return temperature
+
+
+class EssScheme(TemperatureScheme):
+    """
+    Try to keep the effective sample size constant.
```
elbaraim: effective sample size -> effective sample size (ESS)

pyabc/epsilon/temperature.py
```diff
@@ -0,0 +1,502 @@
...
+        weights = np.array(df['w'], dtype=float)
+        pdfs = np.array(df['distance'], dtype=float)
+
+        # compute rescaled posterior densities
+        if kernel_scale == SCALE_LIN:
+            values = pdfs / pdf_norm
+        else:  # kernel_scale == SCALE_LOG
+            values = np.exp(pdfs - pdf_norm)
+
+        # to pmf
```
elbaraim: what is "pmf" ?

pyabc/epsilon/temperature.py
```diff
@@ -0,0 +1,502 @@
...
+            bounds=bounds)
+        beta = ret.x
+
+        temperature = 1. / beta
+        return temperature
+
+
+def _ess(pdfs, weights, beta):
+    """
+    Effective sample size of importance samples.
```
elbaraim: add "(ESS)"

pyabc/acceptor/acceptor.py
```diff
@@ -171,6 +167,20 @@ def __call__(self,
         """
         raise NotImplementedError()
 
+    # pylint: disable=R0201
+    def get_epsilon_config(self, t: int) -> dict:
```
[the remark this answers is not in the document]
yannikschaelte: This is only the abstract class. The implementation is happening only in derived classes. Currently, only the StochasticAcceptor uses it. This function creates a dictionary that contains all the infos from the acceptor at time t that the epsilon will need to propose the next step.

Note: I am not completely happy with how it's implemented now, but it does its job for the moment, however one might like to refactor this at some point.

pyabc/acceptor/acceptor.py
```diff
@@ -287,3 +296,145 @@ def __call__(self, distance_function, eps, x, x_0, t, par):
...
+        pdf_norm = self.pdf_norm_method(
+            kernel_val=kernel.pdf_max,
+            get_weighted_distances=get_weighted_distances,
+            prev_pdf_norm=None if not self.pdf_norms
+            else max(self.pdf_norms.values()))
+        self.pdf_norms[t] = pdf_norm
+
+        logger.debug(f"pdf_norm={self.pdf_norms[t]:.4e} for t={t}.")
+
+    def get_epsilon_config(self, t: int) -> dict:
```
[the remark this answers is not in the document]
yannikschaelte: yes, this is the implementation in the derived class.

pyabc/distance/kernel.py
```diff
@@ -0,0 +1,473 @@
...
+    Parameters
+    -----------
+
+    ret_scale: str, optional (default = SCALE_LIN)
+        The scale of the value returned in __call__:
+        Given a proability density p(x,x_0), the returned value
+        can be either of p(x,x_0), or log(p(x,x_0)).
+
+    keys: List[str], optional
+        The keys of the summary statistics, specifying the order to be used.
```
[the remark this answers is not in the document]
yannikschaelte: Exactly. The problem is that dicts do not specify in what order the keys are traversed. Thus, here the user is supposed to define the keys list, such that the variance and covariance matrix can be constructed based on the order specified here.

Further, `keys` can be less keys than the sumstat dict contains, if not all are to be used here.

pyabc/distance/kernel.py
```diff
@@ -0,0 +1,473 @@
...
+            self.initialize_keys(x_0)
+
+        # compute variance
+        var = self.var(par) if callable(self.var) else self.var
+
+        # difference to array
+        diff = _diff_arr(x, x_0, self.keys)
+
+        # compute pdf
+        log_2_pi = np.sum(np.log(2 * np.pi * var))
```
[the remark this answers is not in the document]
yannikschaelte: What do you mean here? The name ``log_2_pi`` I stole from the numpy implementation of Gaussian distributions ;) . Which product?

pyabc/epsilon/temperature.py
yannikschaelte: good question. The value was found heuristically. However, also other exponents work. I will add a note on this.

pyabc/epsilon/temperature.py
yannikschaelte: probability mass function (i.e. normalized to sum 1)

pyabc/epsilon/temperature.py
```diff
@@ -0,0 +1,502 @@
...
+
+        if t >= self.max_nr_populations - 1:
+            # t is last time
+            temperature = 1.0
+        elif not self.temperatures and self.initial_temperature is not None:
+            if callable(self.initial_temperature):
+                # execute scheme
+                temperature = self.initial_temperature(**kwargs)
+            else:
+                # should be float value
```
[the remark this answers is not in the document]
yannikschaelte: added safety check

pyabc/epsilon/temperature.py
yannikschaelte: The class being abstract? Just that this class does not implement functionality itself, it it just a base class from which all schemes are derived.

pyabc/epsilon/temperature.py
yannikschaelte: Well ... it's difficult to explain in detail here (I just saw this in various instances), but I added a sentence about its usage for initial temperature proposition.

pyabc/epsilon/temperature.py
yannikschaelte: monotonically decreasing, spelled it out.

pyabc/epsilon/temperature.py
yannikschaelte: true. will reformulate.

pyabc/distance/kernel.py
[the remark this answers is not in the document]
elbaraim: using orderedDict is not an option?

pyabc/distance/kernel.py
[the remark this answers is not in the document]
elbaraim: np.log(2 * np.pi * var) --> np.log(2) + np.log(np.pi) +np.log( var)

pyabc/epsilon/temperature.py
elbaraim: thanks :)

pyabc/distance/kernel.py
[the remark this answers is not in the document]
yannikschaelte: Yes, that would be an option. However, then we would need to require the users to use OrderedDicts whenever creating summary statistics, which is also not so nice ...

pyabc/epsilon/temperature.py
elbaraim: oki, otherwise seemed a bit "you have to trust us"

pyabc/epsilon/temperature.py
elbaraim: can you specify? I think I havent seen this abbreviation in the code before.. (if it is already present somewhere, then just keep it)

pyabc/distance/kernel.py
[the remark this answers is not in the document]
yannikschaelte: ah yes, you are right. will do that, since numerically potentially more stable.

pyabc/epsilon/temperature.py
yannikschaelte: will do

pyabc/epsilon/base.py
```diff
@@ -34,15 +36,20 @@ def initialize(self,
 
         t: int
             The time point to initialize the epsilon for.
-
         get_weighted_distances: Callable[[], pd.DataFrame]
             Returns on demand the distances for initializing the epsilon.
+        max_nr_populations: int
+            The maximum number of populations.
+        acceptor_config:
```
EmadAlamoudi: : dict

PR #198, opened 2019-10-23
Release 0.9.19

yannikschaelte:
* Move to cffi>=1.13.1 after that was bug surprisingly quickly fixed (#195).
* Create sub-module for epsilon (#189).
* Add plots for sample and acceptance rate trajectories (#193).


pyabc/epsilon/__init__.py
```diff
@@ -0,0 +1,30 @@
+"""
```
yannikschaelte: Note: The epsilon files were basically just moved.

PR #236, opened 2019-12-19
Feature earlystopping

### LINKED ISSUE #235: Tighter stopping control
Sometimes, there occurs an explosion in the number of samples from one iteration to the next, such that a possible minimum acceptance rate is highly violated. In that case, to not waste computational power it would be of interest to have a tighter control, i.e. to stop the sampling process immediately. This will need to be implemented for each sampler separately (and may not always be possible).

yannikschaelte:
* Implement optional early stopping for the MulticoreEvalParallelSampler when a maximum number of simulations is exceeded.
* Log stopping reason as info.

Fixes #235.

### DIFF
--- a/pyabc/sampler/base.py
+++ b/pyabc/sampler/base.py
@@ -145,8 +145,9 @@ def wrap_sample(f):
     Wrapper for Sampler.sample_until_n_accepted.
     Checks whether the sampling output is valid.
     """
-    def sample_until_n_accepted(self, n, simulate_one, all_accepted=False):
-        sample = f(self, n, simulate_one, all_accepted)
+    def sample_until_n_accepted(
+            self, n, simulate_one, max_eval, all_accepted=False):
+        sample = f(self, n, simulate_one, max_eval, all_accepted)
         if sample.n_accepted != n and sample.ok:
             # this should not happen if the sampler is configured correctly
             raise AssertionError(
--- a/pyabc/sampler/base.py
+++ b/pyabc/sampler/base.py
@@ -14,11 +14,15 @@ class Sample:
     record_rejected: bool
         Whether to record rejected particles as well, along with accepted
         ones.
+    ok: bool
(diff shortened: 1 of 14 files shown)


yannikschaelte:
So far really only implemented for MulticoreEvalParallel and SingleCore. Not straightforward for MultiCoreParticle, but for at least Redis should be straightforward. Left for when needed.
The default behavior has not changed, rather the samplers now got a `check_max_eval` parameter, so they can be configured to stop sampling early. The ABCSMC will then act correctly.

pyabc/sampler/base.py
```diff
@@ -14,11 +15,15 @@ class Sample:
     record_rejected: bool
         Whether to record rejected particles as well, along with accepted
         ones.
+    ok: bool
```
elbaraim: what about rename to, e.g., `succeeded` or `status_flag` or something more informative than `ok`

pyabc/sampler/base.py
```diff
@@ -190,6 +197,7 @@ def sample_until_n_accepted(
             self,
             n: int,
             simulate_one: Callable,
+            max_eval: int = np.inf,
```
elbaraim: is np.inf a type int ?

yannikschaelte: it's the default value, i.e. effectively no limit.

yannikschaelte: yeah, I see your point, ok. it is strictly speaking a float, not an int, but i think that won't matter here...

pyabc/sampler/base.py
yannikschaelte: I used ok (a.k.a. "all correct") as a short intuitive expression (``if sample.ok`` reads better than ``if sample.succeeded``, because it would be the sampler that succeeded, not the sample ..., I would say ...), and as a general flag that can be set to False when something goes wrong. But I can rename if your prefer ...

PR #238, opened 2020-01-07
Redis sec

### LINKED ISSUE #230: Running Redis in protected-mode
Currently, the Redis server is running in pyabc using the flag `protected-mode no`. This may lead to security issues since the Redis instances will be exposed from any external networks. However, according to the Redis recommendation, this flag can be used:

>  only if you are sure you want clients from other hosts to connect to Redis even if no authentication is configured, nor a specific set of interfaces are explicitly listed using the "bind" directive

EmadAlamoudi:
close #230


doc/sampler.rst
```diff
@@ -105,23 +105,62 @@ How to setup a Redis based distributed cluster
...
-main application and by the workers, a Redis server, disabling password
-authentication:
+It is advised to run Redis server in protected mode. Authenticated
+communication will allow only for authenticated access to communicated with
+the server, and reject all unauthorised access. To run Redis server with
+authentication required you need first to modify the redis.conf file.
+
+The redis.conf is the file that contains Redis configuration. Usually,
+it can be found on `/etc/redis/`. to allow safe and secure communication,
+redis.conf file should be reconfigure as follow:
```
yannikschaelte: I think the safest thing for the user will be to copy that file to his home directory, and then only edit that file, and pass it to ``redis-server ``.

doc/sampler.rst
```diff
@@ -105,23 +105,62 @@ How to setup a Redis based distributed cluster
...
 
 .. code:: bash
 
-   redis-server --protected-mode no
+    #bind 127.0.0.1
 
+2. The password authentication must be enabled. To configure that, be sure
+that the masterauth and requirepass configuration options are uncommented, and
+their values are the SAME. Note: it advised to select complex password string
+(e.g., thW0K5tB4URO5a9wsykBH8ja4AdwkQcw).
```
yannikschaelte: I wouldn't give an example here. Users might just copy it, which would make sense very unsecure again ;)

doc/sampler.rst
```diff
@@ -105,23 +105,62 @@ How to setup a Redis based distributed cluster
...
+
+.. code:: bash
+
+    requirepass your_redis_password
+
+3. Finally, you need to restart the Redis server to reload the new configuration.
+
+.. code:: bash
+
+    service redis-server restart
```
yannikschaelte: Do you really need to do the service redis-server etc.? Doesn't it suffice if you first edit the conf file, and then second start the ``redis-server /path/to/redis.conf --port 6379``?

doc/sampler.rst
yannikschaelte: This would also make the part about "How to start a redis server without password" simpler: No need to undo anything, just don't specify that conf file in the redis-server command ...

doc/sampler.rst
EmadAlamoudi: multiple sources require the step of restarting the server. 

see:
1. https://docs.sensu.io/sensu-core/1.3/reference/redis/#redis-master-slave-configuration
2. https://www.digitalocean.com/community/tutorials/how-to-set-up-a-redis-server-as-a-session-handler-for-php-on-ubuntu-14-04#step-2-%E2%80%94-configure-redis-to-accept-external-connections
3. https://community.pivotal.io/s/article/How-to-setup-Redis-master-and-slave-replication

I'm not sure if we can skip it.

yannikschaelte: But I think that only applies when the redis server was running before already, right? Here, we start the server only after the .conf file has been changed. Maybe try once, and if it fails keep it.

Issue #109, opened 2019-05-09
Output number of cores used in MultiCoreSamplers


yannikschaelte:
To inform the user what they are currently doing.

yannikschaelte:
done in #271

PR #263, opened 2020-02-07
N procs

EmadAlamoudi:
closing issue #109


pyabc/smc.py
```diff
@@ -225,6 +225,14 @@ def __init__(self,
         self.min_acceptance_rate = None
 
         self._sanity_check()
+        self.aknowledge_n_procs()
+
+    def aknowledge_n_procs(self):
```
yannikschaelte: maybe this can be done at the end of the pyabc.sampler.multicorebase ``__init__`` function? Would make sense if the check happens in the constructor method of the sampler.

pyabc/smc.py
```diff
@@ -225,6 +225,14 @@ def __init__(self,
...
+        # if sampler is MulticoreEvalParallelSampler, 
+        # then print number of cores
+        if type(self.sampler) is MulticoreEvalParallelSampler:
+            logger.info(f"Total number of used cores is:
```
yannikschaelte: pyABC sometimes uses less cores (and sometimes more, if each simulation is parallelized). Maybe: ``f"Parallelizing the sampling on {self.n_procs} cores."``?

yannikschaelte:
good this finally gets solved, I always ask myself how many cores I am using at the moment :)

pyabc/smc.py
EmadAlamoudi: What do you mean? now the call of the function happened on the __init__ function.

yannikschaelte: I mean not have the function in here, but in pyabc.sampler.multicorebase (and there call it in the class's ``__init__`` function), so that that funcionality is encapsulated.

pyabc/sampler/multicorebase.py
```diff
@@ -37,6 +37,13 @@ def n_procs(self):
         return nr_cores_available()
 
 
+def acknowledge_n_procs(self):
+    # if sampler is MulticoreEvalParallelSampler, 
+    # then print number of cores
+    if type(self.sampler) is MulticoreEvalParallelSampler:
```
EmadAlamoudi: But I'm not sure how to import the `MulticoreEvalParallelSampler` class here

PR #273, opened 2020-02-19
find last id of successful run in the database

EmadAlamoudi:
In case the id was not specified, then pyabc will look for the last ended run.

### DIFF
--- a/pyabc/storage/history.py
+++ b/pyabc/storage/history.py
@@ -210,7 +210,7 @@ class History:
         abcs = self._session.query(ABCSMC).all()
         if len(abcs) > 0:
             for abc in reversed(abcs):
-                if abc.end_time is not None:
+                if len(abc.populations):
                     return abc.id
         return None
 
--- a/pyabc/storage/history.py
+++ b/pyabc/storage/history.py
@@ -209,8 +209,11 @@ class History:
         """
         abcs = self._session.query(ABCSMC).all()
         if len(abcs) > 0:
-            return abcs[-1].id
+            for abc in reversed(abcs):
+                if abc.end_time is not None:
+                    return abc.id
         return None
+
 
     @property
     @with_session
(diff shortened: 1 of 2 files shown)


pyabc/storage/history.py
```diff
@@ -209,9 +209,12 @@ def _find_latest_id(self):
         """
         abcs = self._session.query(ABCSMC).all()
         if len(abcs) > 0:
-            return abcs[-1].id
+            for abc in reversed(abcs):
+                if abc.end_time is not None:
```
yannikschaelte: hm, the end_time is not filled if an abc run is cancelled manually. maybe:

```python
if len(abc.populations) > 0:
    return abc.id
```
but not sure if that is valid code, so check first

yannikschaelte: please also update the docs in line 205ff

EmadAlamoudi: if `abc.populations` contain> 0 but still, the run didn't finish correctly. Then we still can't extract data from the database. 
`df, _ = history.get_distribution(m=m)` will be empty.

yannikschaelte: I think as soon as a population has been written to database, we have to consider it a successful iteration, so len(abc.populations) > 0 would actually be the correct criterion. Usually, users will have only one or two abc analyses in the database, and use different databases for different runs, so usually this should not lead to problems. If one uses one database for multiple runs, it makes sense to note down the id used for a specific run, so that one can then explicitly call ``pyabc.History(..., _id=id)``.

pyabc/storage/history.py
```diff
@@ -203,15 +203,18 @@ def all_runs(self):
...
         """
         abcs = self._session.query(ABCSMC).all()
         if len(abcs) > 0:
-            return abcs[-1].id
+            for abc in reversed(abcs):
+                if len(abc.populations):
+                    return abc.id
         return None
 
+
```
yannikschaelte: remove line, or flake8 will complain ;)

PR #274, opened 2020-02-19
Feature listpopulationsize

nbungi:
Adds ListPopulationSize as populationstrategy. This class was built similar to ListEpsilon.
The abc.run function had to be changed such that it leaves the while loop before preparing the next generation. Otherwise _prepare_next_iteration would be called but no populationsize for the next generation(t+1) could be supplied.


pyabc/populationstrategy.py
```diff
@@ -196,3 +196,30 @@ def adapt_population_size(self, transitions: List[Transition],
 
         logger.info("Change nr particles {} -> {}"
                     .format(reference_nr_part, self.nr_particles))
+
+        
+class ListPopulationSize(PopulationStrategy):
+    """
+    Return PopulationSize values from a predefined list. For every time point
+    enquired later, a PopulationSize value must exist in the list.
+    Parameters
```
yannikschaelte: empty line before parameters (needed for docs parsing)

pyabc/smc.py
```diff
@@ -927,7 +923,14 @@ def run(self,
...
                 logger.info("Stopping: minimum acceptance rate.")
                 break
-
+            elif t==t_max:
+                break
+                
+            # prepare next iteration
+            self._prepare_next_iteration(
+                t+1, sample, population, acceptance_rate)
+
```
yannikschaelte: remove empty line

pyabc/smc.py
```diff
@@ -927,7 +923,14 @@ def run(self,
             elif acceptance_rate < min_acceptance_rate:
                 logger.info("Stopping: minimum acceptance rate.")
                 break
-
+            elif t==t_max:
+                break
+                
+            # prepare next iteration
+            self._prepare_next_iteration(
```
yannikschaelte: I think it makes sense to move the preparation of the next iteration here. good!

test/test_populationstrategy.py
```diff
@@ -108,3 +109,13 @@ def test_transitions_not_modified(population_strategy: PopulationStrategy):
...
     assert same, err_msg
+
+   
+def test_ListPopulationSize():
+    pop_size = pyabc.ListPopulationSize(values=[10, 5, 1.5])
+
+    # might be useful to test integration, but for the moment
+    # standalone tests may suffice
+    assert pop_size(0) == 10
+    assert pop_size(2) == 1.5
```
yannikschaelte: I think calling the population size like this is currently not possible. Instead, nr_particles must be directly used.

yannikschaelte: It is actually planned to have a population size ``__call__`` function which would do exactly this here, which replaces the ``nr_particles``. If you would like to do that for all population strategies, please go ahead, otherwise it would be enough to just make the test run here.

nbungi: Sorry, this was a simple copy paste without thinking... I removed the assertions now, so the test should work.
To be honest I do not understand what you mean with the implementation of a __call__ function so I would leave it with this for now.

test/test_populationstrategy.py
```diff
@@ -108,3 +109,7 @@ def test_transitions_not_modified(population_strategy: PopulationStrategy):
                " modified the transitions".format(population_strategy))
 
     assert same, err_msg
+
+   
+def test_ListPopulationSize():
+    pop_size = pyabc.ListPopulationSize(values=[10, 5, 1.5])
```
yannikschaelte: :) the population size will be rewritten soon, so fine not to have real tests for now.

yannikschaelte:
looks good! I think there might still be an issue with travis, leading to tests not running through. once that is fixed, I will merge this. thanks for the contribution!

yannikschaelte:
changed target to intermediate branch, will fix tests and then forward to develop

yannikschaelte:
@nbungi relocating the "prepare next iteration" causes a bit too much trouble at the moment, thus it will need to remain where it is. I know this introduces some normally not necessary overhead at the end of an iteration, so I will keep in mind to change this at some point, or were there any urgent reasons?

nbungi:
No not really. For now, one will have to keep in mind/document to provide one more population size than needed when using the ListPopulationSize. This was the only reason to relocate the "prepare next iteration".
Alternatively, I guess it would also be possible to append a population size at the end of the given list during initialization such that the user would provide t+1 values.

yannikschaelte:
yes, for the moment I would suggest to just have more population sizes (in general, these will stay the same over time anyway, so it doesn't hurt to just pass ``[pop_size] * 100``

PR #278, opened 2020-03-02
Simplify population size class; implement calibration size

yannikschaelte:


pyabc/populationstrategy.py
```diff
@@ -160,26 +166,34 @@ class AdaptivePopulationSize(PopulationStrategy):
...
-                 *,
-                 max_population_size=float("inf"),
-                 min_population_size=10,
-                 nr_samples_per_parameter=1,
-                 n_bootstrap=10):
-        super().__init__(start_nr_particles,
-                         nr_samples_per_parameter=nr_samples_per_parameter)
+    def __init__(self,
+                 start_nr_particles,
+                 mean_cv=0.05,
```
EmadAlamoudi: space?
```suggestion
                 mean_cv = 0.05,
```

yannikschaelte: right. not possible in python by flake8 standard unless also a type hint is given. but I added one now (that was missing), so now it's better.

Issue #289, opened 2020-04-07
AdaptivePopulation is slow


chrhck:
In my application, determining the new population size can easly take up to two minutes.
I've profiled the bootstrapping function and trace the main chunk of the execution time to the pandas sampling function called here:
https://github.com/ICB-DCM/pyABC/blob/master/pyabc/transition/multivariatenormal.py#L73

I believe this could be drastically sped up, if the sampling was implemented on a numpy array instead of a pandas DataFrame.

(I've recently purged all of pandas from computing heavy parts of my code, because it is _slow_ )

chrhck:
Defining a vectorized rvs for multivariatenormal speeds up the code by a large portion already:
``` def rvs(self, size=None):
        if size is None:
            return self.rvs_single()
        arr = np.arange(len(self.X))
        sample_ind = np.random.choice(arr, size=size, p=self.w, replace=True)
        sample = self.X.iloc[sample_ind]
        perturbed = (sample +
                     np.random.multivariate_normal(
                         np.zeros(self.cov.shape[0]), self.cov,
                         size=size))
        return perturbed
```

chrhck:
An additional speedup can be obtained by vectorizing the `pdf? method of multivariatenormal:

```
python
 def pdf(self, x: Union[pd.Series, pd.DataFrame]):
        x = x[self.X.columns]
        x = x.values
        if len(x.shape) == 1:
            x = x[None, :]
        dens = np.sum(self.normal.pdf(x - self._X_arr) * self.w, axis=0)
        
        return dens if dens.size != 1 else float(dens)
```

EmadAlamoudi:
Dear @chrhck Thanks a lot for your suggestion. Defentatly the use of data type needs further investigation. We might need to look into the use of Panda in the performance bottleneck area.

chrhck:
Yeah, as long as you don't have to to operations on multiple columns at once, a simple `Dict[str, np.ndarray]` is often superior.

chrhck:
Note that the calculation overhead could also easly be reduced by distributing over several workers, similar to the sampling.

yannikschaelte:
fixed in #307 (proposed changes to pdf not implemented but added as comments).

PR #290, opened 2020-04-08
speed improvement by utilizing vectorization and broadcasting

chrhck:
Sped up multivariate normal transition by introducing vectorization/broadcasting.

https://github.com/ICB-DCM/pyABC/issues/289


chrhck:
There is still a small issue in my code: Somewhere the particle weight is transformed into a numpy array of size 1, which causes an error in one of the logging outputs. Still trying to figure out where that happens.

chrhck:
Fixed in https://github.com/chrhck/pyABC/commit/ad76a258fcc84d38a114e9e994f35c32345d6a3c

PR #291, opened 2020-04-08
Multivariatenormal improv

chrhck:
Sped up multivariate normal transition by introducing vectorization/broadcasting.

#289


yannikschaelte:
Dear @chrhck, thanks a lot for this great contribution. We will review and include it as soon as we find time, if that works for you at the moment?

chrhck:
Sure, just one note: the new implementation will require significantly more memory for large population sizes. In my fork if have therefore updated the entire adaptive population calculation to allow for distributed computign via dask. 
For running locally, this could also be fixed by calculating the pdf over slices of the test_X points.

yannikschaelte:
Dear @chrhck , we have now continued this PR in #299. We have taken over your implementation of `rvs()`, which indeed shows considerable performance gains. For `pdf()`, we have so far not taken over your implementation, as the broadcasting leads to memory consumption of GBs for common population sizes. For ease of implementation, we have for the moment kept the old sequential version. You mention that you have distributed this part anyway, so would that be fine for you?

Any comments on #299 are very welcome :) . Thanks again! Yannik

chrhck:
Sounds good: In case you are interested, the distributed implementation is on my fork: https://github.com/chrhck/pyABC/tree/master/pyabc

I had to do quite a large rewrite of the cross-validation helper functions to make this workc so I guess it would be a bigger effort to merge this back into the main repo.

PR #299, opened 2020-05-09
Multivariatenormal improv

### LINKED ISSUE #289: AdaptivePopulation is slow

yannikschaelte:
continued from #291

also: set OMP_NUM_THREADS.

fixes #289, #302


pyabc/transition/util.py
```diff
@@ -1,9 +1,10 @@
 import numpy as np
 
 
-def smart_cov(X_arr, w):
-    """
-    Also returns a covariance of X_arr consists of only one single sample
+def smart_cov(X_arr: np.ndarray, w: np.ndarray) -> np.ndarray:
+    """Create sample covariance matrix.
```
EmadAlamoudi, line 5: """ \n

PR #303, opened 2020-05-11
Use setup-R on github actions

yannikschaelte:


yannikschaelte:
closing as currently not worth the effort of setting up. once "ubuntu-latest" gets updated, we may run into errors and then need to update the R source manually, which should be fine.

Issue #202, opened 2019-10-23
Sample numbers plots normalize by ESS


yannikschaelte:
Add an option to normalize the sample numbers and acceptance rates plot by the ESS.

Havaldar:
@yannikschaelte I was hoping to tackle this issue and just wanted to make sure no one was already on it. Also would normalizing the sample numbers be equivalent to the relative ESS plot or is it N / ESS? (I am really sorry if this is a silly question)

yannikschaelte:
Hi @Havaldar , I think no-one is working on this yet -- contributions are very welcome ;) .

For the acceptance rate trajectory, it would be not to plot population_size / n_samples, but ESS / n_samples. This would allow assessing how many simulations were needed per effectively accepted particle, which is something one is often interested in. Similarly, for the sample numbers (stacked bar and time series) plots, it would be not plotting n_samples, but n_samples * population_size / ESS. 

However, on second thoughts I think that the latter may be kind of superfluous, as one can see the impact of samples per effective acceptance already in the acceptance rate plot, thus I think only the first one should suffice ...

PR #345, opened 2020-09-08
Normalize Acceptance Rate Plot by ESS

Havaldar:
Tackles issue #202 

- Defaults to plotting acceptance rate by population size
- Uses flag to replace population size by effective sample size

Example plot (with acceptance rate default and ESS plot)
example

### DIFF
--- a/pyabc/visualization/sample.py
+++ b/pyabc/visualization/sample.py
@@ -5,6 +5,7 @@ from typing import List, Union
 
 from ..storage import History
 from .util import to_lists_or_default
+from ..weighted_statistics import effective_sample_size
 
 
 def plot_sample_numbers(
@@ -263,7 +264,8 @@ def plot_acceptance_rates_trajectory(
         yscale: str = 'lin',
         size: tuple = None,
         ax: mpl.axes.Axes = None,
-        colors: List[str] = None):
+        colors: List[str] = None,
+        normalize_by_ess: bool = False):
     """
     Plot of acceptance rates over all iterations, i.e. one trajectory
     per history.
@@ -288,6 +290,9 @@ def plot_acceptance_rates_trajectory(
         The size of the plot in inches.
     ax: matplotlib.axes.Axes, optional
         The axis object to use.
+    normalize_by_ess: bool, optional (default = False)
+        Indicator to use effective sample size for the acceptance rate in
+        place of the sample size.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


pyabc/visualization/sample.py
```diff
@@ -312,10 +317,18 @@ def plot_acceptance_rates_trajectory(
         # note: the first entry of time -1 is trivial and is thus ignored here
         h_info = history.get_all_populations()
         times.append(np.array(h_info['t'])[1:])
-        samples.append(np.array(h_info['samples'])[1:])
         pop_sizes.append(np.array(
             history.get_nr_particles_per_population().values[1:]))
 
+        if normalize_by_ess:
```
yannikschaelte, line 323: I think there is one error here: Not samples should be replaced by ESS, but pop_sizes should be replaced by ESS, t.t. the resulting rations are usually <= 1. So, just move this code block up a bit ;) Sorry, if I miscommunicated that.

yannikschaelte: Please also update the docstring accordingly.

yannikschaelte (changes requested):
Hi, that looks (except one thing) really nice. Could you also add a test case to https://github.com/ICB-DCM/pyABC/blob/master/test/visualization/test_visualization.py#L53? I.e. just run the same code with the new option enabled. I know that this is not a proper unit test, but for visualization that is all we currently do.

pyabc/visualization/sample.py
Havaldar: done!

Havaldar:
made all the changes! (Thank you so much for the review :) )

test/visualization/test_visualization.py
```diff
@@ -74,7 +74,8 @@ def test_acceptance_rates_trajectory():
         histories, labels, yscale='log', rotation=76)
     _, ax = plt.subplots()
     pyabc.visualization.plot_acceptance_rates_trajectory(
-        histories, labels, yscale='log10', rotation=76, size=(10, 5), ax=ax)
+        histories, labels, yscale='log10', rotation=76, size=(10, 5), ax=ax
+        normalize_by_ess=True)
```
yannikschaelte, line 78: One last thing: Could you once run this with normalize_by_ess=False, and once with True? So as to test both modes.

yannikschaelte (approved):
Nice, thanks! 

(And in your plot I see that we need to enforce integer values on the x axis when plotting populations. Not sure how to best do that -- needs to be integer, but should not be _all_ integers in case many time points exist. Leaving that for future work :) .)

test/visualization/test_visualization.py
Havaldar: Done!

yannikschaelte:
Great! As the tests do not seem to run, I will change base and from there do a PR do develop.

PR #348, opened 2020-09-28
References

elbaraim:
added reference list using pyABC


references.rst
```diff
@@ -0,0 +1,30 @@
+References
+==========
+
+List of publications using pyABC, total amounting to 28 citations.
```
yannikschaelte, line 4: Number needs to be updated (should be 16 if I can count).

references.rst
```diff
@@ -0,0 +1,30 @@
...
+If you applied pyABC in your work, please notify us through a new Github issue.
```
yannikschaelte, line 5:
```suggestion
If you applied pyABC in your work, please consider notifying us.
```
(some may prefer to just write an e-mail).

yannikschaelte (approved):
Can you link to this from the documentation? Maybe in https://github.com/ICB-DCM/pyABC/blob/master/doc/index.rst just add under "about.rst".

elbaraim:
> Can you link to this from the documentation? Maybe in https://github.com/ICB-DCM/pyABC/blob/master/doc/index.rst just add under "about.rst".

i do not understand what you mean, do not know how to properly include it

yannikschaelte:
> > Can you link to this from the documentation? Maybe in https://github.com/ICB-DCM/pyABC/blob/master/doc/index.rst just add under "about.rst".
> 
> i do not understand what you mean, do not know how to properly include it

@elbaraim I added it now (and I made it look overly complicated by at the same time refactoring index.rst).

PR #357, opened 2020-11-11
Enable "look-ahead" sampling for adaptive components

yannikschaelte:
In this case the final decision criterion is not known, therefore the checking for acceptance is delayed. In the given implementation, it is performed on the main thread as soon as all required data are available.


yannikschaelte:
due to merge conflicts, this has been moved to #374 and now done

PR #362, opened 2020-11-23
update to petab test suite changes

yannikschaelte:


FFroehlich (approved):
Looks good from my end, but I think @dweindl is more familiar with the code

dweindl (approved):

PR #370, opened 2020-11-28
Add maximum walltime stop condition

### LINKED ISSUE #369: Time as a stopping condition
Hello Guys!
Great work with pyABC!
One (small I hope) request. For some reason, I need to define the stopping condition as the time amount. I have implemented it comehow by myself but there is a problem with installing it in my jupyter notebook, upgrading up to new version and so. So the question and big big request is: would it be possible to add such a feature (adding the stopping condition as a time (in seconds)) in the official release?
Thank you in advance!!
Leszek

yannikschaelte:
Fixes #369

### DIFF
--- a/pyabc/inference/smc.py
+++ b/pyabc/inference/smc.py
@@ -1,10 +1,10 @@
 """ABC-SMC"""
 
-import datetime
 import logging
 from typing import List, Callable, TypeVar, Union
 import numpy as np
 import copy
+from datetime import datetime, timedelta
 
 from pyabc.acceptor import (
     Acceptor, UniformAcceptor, SimpleFunctionAcceptor, StochasticAcceptor)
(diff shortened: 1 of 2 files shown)


lesiwik:
Is there a distinction between time-based and evaluation-based stopping? or missing time means evaluation-based stopping and vice versa?
could you pls provide sample calls?

yannikschaelte:
Basically, the run() method looks as follows:

```python
def run(self,
            minimum_epsilon: float = None,
            max_nr_populations: int = np.inf,
            min_acceptance_rate: float = 0.,
            max_total_nr_simulations: int = np.inf,
            max_walltime: timedelta = None) -> History:
```

By default, no stopping criterion is set, meaning pyabc will do more iterations until manually interrupted. If you only want to specify a simulation budget of say 100,000 model simulations, you can call `run(max_total_nr_simulations=1e5)`, if you only want to specify a maximum wall clock time, `run(max_walltime=timedelta(hours=23)`. If you specify multiple criteria, pyABC will stop as soon as any is hit.

lesiwik:
Yanik, Could you pls confirm that checking the number of evaluation after every generation is merged into the latest build?

yannikschaelte:
Yes, it is in pyabc 0.10.9, https://github.com/ICB-DCM/pyABC/blob/master/pyabc/inference/smc.py#L679. Does it not work for you?

PR #374, opened 2020-12-06
Very parallel sampling: The adaptive way

yannikschaelte:
* Allow using adaptive things [distances, epsilons, ...] in look-ahead mode via delayed evaluation
* Also allow adaptive population sizes (fixes #357)
* Log relevant numbers to csv in redis server.
* Correctly handle
* Allow redis workers to fail on simulation error for testing purposes (optional, command line arg).

TBD at some point:
* Recognition of the calibration iteration (which has weird timing t=0 and does some funny stuff) can be improved (currently via `all_accepted`).

What does not work (any never will with the current implementation):
* Integrated simulation and evaluation.


pyabc/sampler/redis_eps/cli.py
```diff
@@ -65,12 +70,11 @@ def get_int(var: str):
     ana_id = analysis_id
```
FelipeR888, line 70: So this is  what you added to avoid several runs to somehow overlap?

yannikschaelte: exactly. all variables are now id'd via analysis_id and time index, so the association should be unique.

pyabc/sampler/redis_eps/cli.py
```diff
@@ -44,6 +48,7 @@ def handle(self, *args):
 def work_on_population(analysis_id: str,
                        t: int,
                        redis: StrictRedis,
+                       catch: bool,
```
FelipeR888, line 51: What does this catch variable do?

yannikschaelte: that is mostly unrelated to the look-ahead thing, during tests I just wanted a way to make the workers fail directly when something goes wrong. the thing is, currently (without catch=False) when using the no-acceptance-delay version of look-ahead with some adaptive distance etc, there will never be thrown an error, but the sampler will just fail many times at at some point switch to non-look-ahead mode then.

this "catch" is however only semi-useful, because the calling main process never gets notified, one still has to manually check.

FelipeR888 (approved):

PR #379, opened 2020-12-09
Implement static redis sampler

yannikschaelte:
Note: The dynamic sampler was not touched, apart from adding a "MODE" flag. It was just revamped in order to exploit synergies of both dynamic and static sampler.

### DIFF
--- a/pyabc/sampler/redis_eps/sampler.py
+++ b/pyabc/sampler/redis_eps/sampler.py
@@ -1,3 +1,5 @@
+"""Redis based sampler base class and dynamic scheduling samplers."""
+
 import numpy as np
 import pickle
 from time import sleep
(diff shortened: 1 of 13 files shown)


pyabc/sampler/redis_eps/sampler.py
```diff
@@ -270,37 +319,19 @@ def clear_generation_t(self, t: int) -> None:
...
+         .delete(idfy(N_ACC, ana_id, t))
+         .delete(idfy(N_REQ, ana_id, t))
+         .delete(idfy(N_FAIL, ana_id, t))
+         .delete(idfy(ALL_ACCEPTED, ana_id, t))
+         .delete(idfy(N_WORKER, ana_id, t))
+         .delete(idfy(BATCH_SIZE, ana_id, t))
+         .delete(idfy(IS_LOOK_AHEAD, ana_id, t))
+         .delete(idfy(MODE, ana_id, t))
+         .delete(idfy(QUEUE, ana_id, t))
+         .execute())
```
FelipeR888, line 334: How come the only thing happening in the sample_until_n_accepted function is adding and deleting to the pipeline?

FelipeR888: Nevermind, i just missed the button to open the rest ^^

FelipeR888: But still, all those pipeline.set and pipeline.delete parts used to be in a separate function, didnt they?

FelipeR888 (approved):

pyabc/sampler/redis_eps/sampler.py
yannikschaelte: They are actually. GitHub just displays the difference very weirdly. They should be in start_generation_t and clear_generation_t functions if I remember correctly.

PR #385, opened 2020-12-17
Add pre-commit to ensure a consistent style

basnijholt:
I have added
- pyupgrade
- black
- isort
- flake8
- some end of file and trailing whitespace checks
- CI check that runs pre-commit


basnijholt:
I've just done some cherry-picking to change the base from `master` to `develop` after reading https://github.com/ICB-DCM/pyABC/pull/384#issuecomment-747339878.

basnijholt:
The CI passes on my fork: https://github.com/basnijholt/pyABC/actions/runs/427816772

yannikschaelte:
Hi @basnijholt , thanks, this looks really nice! I had in the past thought about fixing all extended pep8 style issues, however so far decided against that, as there are multiple open PRs, also some detached ones at collaboration partners, still open at the moment. Merging the style enforcements here would likely lead to trouble on that branches, in particular with linear history, therefore I would rather not have that at the moment ... maybe when most project branches have been merged (hopefully this year at some point) may be a good time to tidy things up.

So, if I see correctly those pre-commit hooks (in particular black) perform those code formattings? It seems to do a pretty decent job, better than what I know from R. It might be good to define a toc environment for the flake8 checks, to install all checks we want to check against (flake8-print, ... --> setup.cfg[quality]) in there specifically.

yannikschaelte:
Unfortunately it would at the moment be too much effort to include this here, as I disagree with some of black's suggestions, and I am not sure whether the massive reformatting will be a problem with detached branches. We'll close for the moment, but keep it in mind!

PR #387, opened 2020-12-31
Speed up parameter handling

yannikschaelte:
Speed up parameter density evaluation in the transitions.
* Avoid conversions from and to pd.Series (which is highly costly) by exporting pyabc.Parameter entities in .rvs() and accepting them as input in .pdf().
* Particularly in MultivariateNormalTransition and LocalTransition .pdf(), directly convert Parameter to np.ndarray by key iteration (could be improved if needed). This also ensures that the same order of keys is maintained in covariance matrices and samples.
* In .rvs(), use numpy.random.choice instead of pandas.DataFrame.select for speed-up.

Observed speed-up for very low runtime models: roughly 30%, see also #384.

Further:
* Tidy up some transitions.pdf() and test_* code.
* Add some benchmark study logs in test_performance on critical code parts.
* When multiple parameters are passed to .pdf(), this is expected as a pandas.DataFrame. I would expect the overhead to be ok here. Alternatively, List[Parameter] would be feasible. Only used in the adaptive population size atm afaik.
* Passing pd.Series stays allowed for legacy.
* Fix pandas <1.2.0 until https://github.com/PEtab-dev/PEtab/pull/493 is fixed.


yannikschaelte:
Note: Merging must wait a little until a currently running sampler comparison study has finished (2-3 weeks max. probably).

PR #398, opened 2021-01-20
Web server 2

EmadAlamoudi:
a Dash webserver to visualize pyABC output directly from the database file.


EmadAlamoudi:
I think that the remaining of flake8 issues are either can be ignored, e.g., `S108 Probable insecure usage of temp file/directory` or from other files

yannikschaelte:
@EmadAlamoudi is this ready for review? I think some tests would be good, e.g. setting up a server. I think this is hard to test as it is a GUI, but adding a test case that at least once starts and stops a server would be good.

EmadAlamoudi:
It seems that it is easy to start the server, but then how to close it? another thread needs to handle killing the server. should I go for that?

yannikschaelte:
> It seems that it is easy to start the server, but then how to close it? another thread needs to handle killing the server. should I go for that?

I think yes. And the test would "pass" if the server starts successfully.

EmadAlamoudi:
Done.

EmadAlamoudi:
hmm, that does not seems to be useful. The test always successful even if there is an actual error. I will have another look.

EmadAlamoudi:
testing is done. Please have a look.

doc/examples/quickstart.ipynb
```diff
@@ -0,0 +1,416 @@
+{
```
yannikschaelte, line 1: this notebook does no longer exist, or has been renamed model_selection.ipynb

yannikschaelte:
Ok: Looks overall nice already. Some comments:
* I would go for a "light/white" theme, as the plots will have a white background, and it is too much of a contrast otherwise. Also the pyABC logo does not work with black.
* Note that there are .svg+.png+etc. files for the logo already in doc/logo/...
* I would suggest to always use a default. E.g. when selecting the database, the default run should be the last. When selecting a parameter plot, the default should be the first parameter (or the one previously shown), etc., s.t. one does not always have to manually do "Select ... -> Parameter name".

Errors I ran into:
* When using a database with a single run, it did not do anything when clicking on the run selection, the right part just stays black.
* When plotting multiple plots in the "Runs plots" section, it gave me errors like `RuntimeError: main thread is not in main loop
`. That seems to indicate sth is wrong with the underlying logic ...
* Flake8 cannot pass yet. When everything runs, please consider the changes I recently made, as explained [here](https://github.com/ICB-DCM/pyABC/blob/develop/CONTRIBUTE.rst) to use `pre-commit install` and `tox -e flake8` to check the code locally already.

yannikschaelte:
I think it woud also be good to just have one web server. If this one replaces the old implementation, all functionality should be copied over.

EmadAlamoudi:
* I think all the functionality of the previous server is already included. do you anything in particular?

PR #401, opened 2021-01-23
Various improvements

yannikschaelte:
* Add ground truth to noise notebook
* Rename notebook "quickstart" -> "model_selection"
* Split notebook tests in 2, extend, reduce test matrix (python 3.7)
* Improve History output at the beginning and end (e.g. no end_time at beginning; show duration)
* Add walltime plots (`plot_total_walltime`, `plot_walltime`, `plot_eps_walltime`)
* Make sure `ABCSMC.run()` is always properly finished by wrapper


pyabc/storage/history.py
```diff
@@ -369,48 +360,47 @@ def get_all_populations(self):
...
             The epsilon represented as json string.
-
-        population_strategy_json_str: str
+        population_strategy_json_str:
             The population strategy represented as json string.
-
+        start_time:
+            Start time of the analysis.
         """
+        if start_time is None:
```
EmadAlamoudi, line 397: why do we need this check? Does that mean the user can define the start time? if yes, why?

pyabc/visualization/epsilon.py
```diff
@@ -1,56 +1,53 @@
...
     Plot epsilon trajectory.
 
     Parameters
     ----------
-
-    histories: Union[List, History]
+    histories:
         The histories to plot from. History ids must be set correctly.
-    labels: Union[List ,str], optional
+    labels:
```
EmadAlamoudi, line 27: why the datatype + default value are removed?

EmadAlamoudi (approved):

pyabc/storage/history.py
yannikschaelte: valid question. The user cannot define it, but I thought it is more transparent if that time is defined in the ABCSMC class, and not implicitly in the History class. Because then whoever changes the code knows that putting that function to another place may affect the time that is saved in the history.

pyabc/visualization/epsilon.py
yannikschaelte: We have datatype annotations now always (mostly) directly in the python code. Having it in the docstring as well is thus superfluous, and potentially leads to problems if both are not synchronized.

EmadAlamoudi: totally agreed.

FelipeR888 (approved):

PR #403, opened 2021-01-28
Allow pickling of AMICI models

lcontento:
This PR aims to make pyABC models created from PEtab/AMICI to be pickable, enabling distributing the computation across nodes.

This follows closely the implementation in pyPESTO. If the user makes modifications to the model object or provides a model object obtained from a different PEtab problem, then the pickling will not work. But I guess this is not very probable.

We could make the `PEtabImporter` only accept a petab problem as an argument solving this problem. The user could still set solver options by accessing the returned model object.

In order for the workers to reuse the AMICI module compiled by the master process I had to make a small change in AMICI https://github.com/AMICI-dev/AMICI/pull/1383

### DIFF
--- a/pyabc/model.py
+++ b/pyabc/model.py
@@ -239,7 +239,11 @@ class SimpleModel(Model):
                  sample_function: Callable[[Parameter], Any],
                  name: str = None):
         if name is None:
-            name = sample_function.__name__
+            if hasattr(sample_function, '__name__'):
+                name = sample_function.__name__
+            else:
+                # sample_function is a callable object
+                name = sample_function.__class__.__name__
         super().__init__(name)
         self.sample_function = sample_function
 
--- a/pyabc/petab/amici.py
+++ b/pyabc/petab/amici.py
@@ -2,6 +2,7 @@ import logging
 from collections.abc import Sequence, Mapping
 from typing import Callable, Union, Dict
 import copy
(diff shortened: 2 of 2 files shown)


yannikschaelte (approved):
Looks perfect! I will add some docs before merging to master, but fine as it is.

yannikschaelte:
I added you as contributor, so you can base your PR from here. If it's time-critical, I can just force-merge here and check tests (lint) afterwards.

lcontento:
> I added you as contributor, so you can base your PR from here. If it's time-critical, I can just force-merge here and check tests (lint) afterwards.

Maybe it's better if I clean it up and add docs before merging (this is exactly the kind of thing that I am going to forget about if I postpone it). I can use my branch for KoCo19 for now (like I already do for PEtab, AMICI and pyPESTO...).

EmadAlamoudi (approved):
looks great!

yannikschaelte:
There have been changes in the meantime to the relevant files, thus probably merging will be necessary.

yannikschaelte:
@lcontento do you intend to continue here? Then calling the formatting pre-commit hooks on your fork would be necessary. Otherwise also happy to look into enabling pickling for amici models.

yannikschaelte:
@lcontento copied now over to #527 due to merge issues

PR #410, opened 2021-02-10
Fix redis for later started workers

yannikschaelte:
Due to some weird behavior of the redis server, initially only posting a 1 instead of the start message, in that case also check whether an analysis setting seems to be there.


yannikschaelte:
@EmadAlamoudi @FelipeR888 I guess this may explain also your issues

FelipeR888 (commented):
So now it shouldnt be a problem anymore if the python program is started before any worker connects to the server?

yannikschaelte:
Yes, exactly.

PR #413, opened 2021-02-12
Improve PEtab support

yannikschaelte:
* Fix wrong import of log distributions (wrong arguments, wrong scale before). Principle: Prior returns parameters on prior scale, therefore model has to rescale inputs.
* Add tests for prior generation
* Discard that overly complicated allowing to estimate fixed parameters too
* Now, pyABC should support all priors defined in PEtab.
* Note: Currently sampling is always performed on prior scale, which should be statistically correct , but ignores the PEtab idea of defining a separate parameter scale. How that is to be interpreted in terms of transformation is yet TBD.

### DIFF
--- a/test/petab/test_petab.py
+++ b/test/petab/test_petab.py
@@ -1,5 +1,6 @@
 import os
 import numpy as np
+import scipy.stats
 import pandas as pd
 import pytest
 import git
@@ -55,6 +56,9 @@ def prior_specs(request):
 
 def test_petab_prior(prior_specs):
     """Test whether the prior is correctly defined by sampling from it."""
+    # need to fix random seed due to stochastics of multiple testing
+    np.random.seed(0)
+
     # extract settings
     scale, prior_type, var1, var2, parameter_df = prior_specs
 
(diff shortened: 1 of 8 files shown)


test/petab/test_petab.py
```diff
@@ -1,13 +1,173 @@
...
+    # -- MEAN AND VARIANCE -- #
+
+    # sample mean and variance
+    mean = np.mean(samples)
+    var = np.var(samples)
+
+    # ground truth mean and variance
+    if prior_type in [C.UNIFORM, C.PARAMETER_SCALE_UNIFORM]:
+        mean_th = var1 + (var2 - var1) / 2
+        var_th = (var2 - var1)**2 / 12
```
dweindl, line 90: Where does the `12` come from?

dweindl (approved):
Not feeling overly confident, but looks good to me.

test/petab/test_petab.py
yannikschaelte: that's the variance of  (a,b) uniforma variable https://en.wikipedia.org/wiki/Continuous_uniform_distribution

dweindl: I guess one should know, thanks :see_no_evil:

LeonardSchmiester (approved):
Same as for Daniel regarding confidence level. But looks fine from what I can tell.

PR #418, opened 2021-02-19
fix LA sampler weighting

yannikschaelte:
Employ subpopulation-wise self-normalization.


EmadAlamoudi (approved):

FelipeR888 (approved):
So for now the proposal_id is always directly set to either 0 (for preliminary) or -1 (for final)?
Ill have to keep that in mind if i still want to implement a continuously updated to the preliminary, but for now this looks good.

PR #420, opened 2021-02-20
Turn summary statistics callable into class

yannikschaelte:


yannikschaelte:
this test balloon has popped, not gonna pursue

PR #428, opened 2021-02-22
Limit look-ahead sample number in delayed mode

yannikschaelte:


.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -13,7 +13,7 @@ assignees: ''
 **Expected behavior**
```
EmadAlamoudi, line 13: capital maybe?

EmadAlamoudi (approved):

.github/ISSUE_TEMPLATE/bug-report.md
yannikschaelte: Changed it to lower-case to make it fit with the other titles :laughing:

pyabc/sampler/redis_eps/sampler.py
```diff
@@ -169,12 +176,14 @@ def __init__(self,
                  batch_size: int = 1,
                  look_ahead: bool = False,
                  look_ahead_delay_evaluation: bool = True,
+                 max_n_eval_look_ahead_factor: float = 10.,
```
FelipeR888, line 179: Wouldnt it maybe be better to default it to infinity, since this is kind of only useful when memory is an issue? 
Also, if we ever want to enable the buffering for several generations ahead.

FelipeR888 (approved):

pyabc/sampler/redis_eps/sampler.py
yannikschaelte: Not sure. I think this may also be an issue with the short-running models, where we basically had a few thousand evaluations in round two, all of which were then accepted nearly, at a population size is 100 maybe. This was there because the time between iterations was so long. Then checking all acceptances may again take some time, so that setting it to inf may indeed result in longer times than a reasonable limit. Not sure if 10 is too aggressive? Are there cases where acceptance rate drops by a factor of 10 quickly? Then one could use a larger default ...

PR #448, opened 2021-05-08
Add dynamical mode that only waits for relevant particles in redis server

yannikschaelte:
Previously, after a generation the process always waited for all workers to finish, i.e. also for simulations that were started after the last accepted one, which is unnecessary, as these are disregarded anyway. This is now no longer done. When the new generation starts, typically not all workers will be available from the beginning, but gradually join. Both modes are supported via the `wait_for_all_samples` flag (new is default).

Gains: Speed-up, tbd.

Currently only supported by the Redis server, as for the other implementations, management of loose processes must be defined. Cancellation of running simulations would also be possible via monitored subprocessing, was however disregarded for the moment as this could lead to uncleanly terminated processes.


pyabc/sampler/redis_eps/work.py
```diff
@@ -130,12 +129,16 @@ def get_int(var: str):
 
         # check if in look-ahead mode and should sleep
         if is_look_ahead and get_int(N_EVAL) >= max_n_eval_look_ahead:
-            # sleep 0.1 seconds
+            # sleep ... seconds
```
FelipeR888, line 132: why ... when the 0.1 is still hard-coded?

yannikschaelte: just for fun :) just wanted to indicate it's seconds, and when changing the 0.1 at some point not forget to change it twice

pyabc/sampler/redis_eps/util.py
```diff
@@ -1,5 +1,13 @@
 import signal
 import sys
+import cloudpickle as pickle
+from redis import StrictRedis
+
+from .cmd import ACTIVE_SET, idfy
+from .redis_logging import logger
+
+# lock for accessing the active set data structure
+ACTIVE_SET_LOCK = "active_set_lock"
```
FelipeR888, line 10: Why is this here and not in the cmd.py file?

yannikschaelte:
Here is a very preliminary evaluation:

On the tumor model (15 iterations, population size 200, 200 cores, median epsilon), it gave on a single run a speed-up of 20-30%, denoted DYN2 in the plot. On this run, LA did not improve, but as seen before, runtimes are quite heterogeneous, to be investigated in more detail.

image

Relative runtimes: [100.        ,  72.50725898,  77.76705374, 235.32882309]

On an ODE model with 0.25 var lognormal time variance, 10 repetitions, DYN vs DYN2 did not show major differences, here LA gave the biggest advantage. Combining DYN2+LA was fastest overall, with another 5% speed-up.

image

Relative mean runtimes: [100.        ,  98.9614077 ,  62.6626877 ,  57.85047793, 159.61560853]

pyabc/sampler/redis_eps/util.py
yannikschaelte: yeah, probably makes sense to define in cmd.py all redis string constants.

pyabc/sampler/redis_eps/sampler.py
```diff
@@ -636,3 +660,19 @@ def self_normalize_within_subpopulations(sample: Sample, n: int) -> Sample:
...
+def _log_active_set(
+    redis: StrictRedis, ana_id: str, t: int, id_results: List[Tuple],
+    batch_size: int,
+) -> None:
+    """Log the status of active simulations after the first n acceptances."""
+    accepted_ids = [id_result[0] for id_result in id_results]
+    active_set = get_active_set(redis=redis, ana_id=ana_id, t=t)
+    # remove entries that are already accepted (runtime conditions)
+    active_set = active_set.difference(accepted_ids)
+    earlier = {ix for ix in active_set if max(accepted_ids) > ix}
```
FelipeR888, line 674: So this "earlier" is the set we want to wait for, right? I dont see where you use that.

yannikschaelte: This function here is only for logging. The collection is checked in

https://github.com/ICB-DCM/pyABC/pull/448/files#diff-2f41be0995fef5a262d8bf363d29590d9bec59065f8d219a2431ce40ce589344R263

There, it is checked whether it contains an entry with a smaller index than the maximum index of all so far accepted particles. Because it is those particles that could potentially, if accepted, replace that maximum index particle or others in the final accepted population.

FelipeR888: Right; saw it now. Should work imo :+1:

FelipeR888 (approved):

PR #529, opened 2021-12-04
Remove external handler cleanup as not used + correctly implemented

yannikschaelte:


pyabc/external/base.py
```diff
@@ -95,19 +89,6 @@ def create_loc(self):
                 suffix=self.suffix, prefix=self.prefix, dir=self.dir
             )[1]
 
-    def clean_simulation_output(loc):
```
yannikschaelte, line 98: static function? then would need to be indicated.

PR #608, opened 2023-07-06
Fix PEtab test suite

### LINKED ISSUE #607: PEtab test suite failing
**Bug description**
Due to currently unknown changes in petab-test-suite or libpetab-tests or petab, most tests in the test suite are failing for various reasons. Currently, we can thus not be sure whether pyABC satisfies the most recent PEtab standard.

dweindl:
PEtab test suite started failing after moving from amici==0.17.1 -> amici==0.18.0, due to importing different models with the same name. Fixed by using distinct names.

Adapt to `petabtests` API changes.

Fixes #607


yannikschaelte (approved):
Thank you very much, awesome! It would have taken me some time to guess that the diverse errors were caused by the wrong model name being used.

Complemented with fixes of multiple other things that failed, now ready to merge.

PR #616, opened 2023-11-06
Fix assets for pyabc visserver.

stephanmg:


stephanmg:
Should create also a new patch release soon.

PR #642, opened 2024-10-29
Update deploy.yml

stephanmg:


arrjon (approved):
let's see if that works, I also updated the needed tests so that they align now with the newer python versions

PR #665, opened 2026-05-28
Add Python 3.14 to CI 

kilianvolmer:
To check for compatibility with the most recent python version I suggest adding checks for Python 3.14 to the CI.


arrjon:
Thank you for pointing to this problem. I will look into it

PR #668, opened 2026-07-24
Add better time tracking

arrjon:
This pull request adds detailed execution time profiling to the ABC-SMC run, including both wall-clock and fine-grained sub-step timings. Additionally, per-generation walltimes are now stored in the database, and visualizations have been updated to reflect these more accurate timings.

I also cleaned up some minor code.

Resolves #325 and #667

Issue #47, opened 2019-01-29
Normalization of weights


yannikschaelte:
Before being saved into the database, the weights of particles are normalized s.t. all particles belonging to one model have a summed weight of 1. However, there are multiple problems:
1. When there are multiple simulations per particle (as encoded in population_strategy.py()), the sum of the normalized weights of all summary statistics (e.g. obtained via history.get_weighted_sum_stats()) will in general not sum to 1. This must be taken into account when e.g. computing effective sample sizes in transition.multivariatenormal.py
2. Only the normalized weights and the model probabilities are stored in the database. Thus, the original weights cannot be reproduced afterward, unless also e.g. the total model weights are stored.

Solution:
1. When retrieving e.g. weights + sum stats, always renormalize to sum = 1, or make sure in all using methods that they check ``w = w/sum(w)``. Then the analyses should be right, I guess.
2. Probably doesn't matter, since we won't need that information, or possibly can retrieve it via the transition kernel.

Maybe it would also make sense to not normalize the weights that are inserted into the database at all. Then it should be easiest to always apply the correct normalization depending on what's needed.

yannikschaelte:
One has to be careful with using pyabc with databases previously generated then, so this would be a second order version update, and at the same time version checks in the database will become necessary.

PR #670, opened 2026-07-24
Particle weights are now stored using the global normalization

arrjon:
I took the chance that #668 changes the database anyway to resolve the old issue  #47. Particle weights are now stored using the global normalization (weights sum to 1 across all particles of all models, matching the in-memory ``Population`` representation). This is bundled into database version 2 and handled by ``abc-migrate`` for existing databases. 

Moreover, the PR cleans up some more type hints, solves some minor bugs across the codebase and adds some more tests.


arrjon:
Wait for #668 to be merged before

arrjon:
Ready to be merged when test pass

pyabc/visualization/kde.py
```diff
@@ -247,8 +247,6 @@ def plot_kde_1d(
     if ax is None:
         _, ax = plt.subplots()
     ax.plot(x_vals, pdf, **kwargs)
-    # TODO This fixes the upper bound inadequately
-    # ax.set_ylim(bottom=min(ax.get_ylim()[0], 0))
```
kilianvolmer, line 251: So we do not want to set a y_lim?

pyabc/storage/migrations/versions/2_20260724_add_populations_wall_time.py
```diff
@@ -21,7 +29,16 @@ def upgrade():
...
         column=sa.Column('wall_time', sa.FLOAT, nullable=True),
     )
+    # per-model-normalized weights -> global weights (w := w * p_model)
+    op.execute(
+        'UPDATE particles SET w = w * ('
+        'SELECT p_model FROM models WHERE models.id = particles.model_id)'
+    )
 
 
 def downgrade():
```
kilianvolmer, line 43: Shouldn't there then also be a downgrade option for the other database change?

test/base/test_storage.py
```diff
@@ -282,6 +282,86 @@ def test_sum_stats_save_load(history: History):
...
+    )
+
+    # model probabilities are the per-model sums of the global weights
+    mp = history.get_model_probabilities(t=0)
+    assert np.isclose(mp.loc[0, 'p'], 0.4)
+    assert np.isclose(mp.loc[1, 'p'], 0.6)
+
+    # the RAW stored weights are global and sum to 1 (per-model storage would
+    # have summed to the number of models, i.e. 2)
+    ext = history.get_population_extended(t=0, tidy=False)
```
kilianvolmer, line 333: What is this supposed to test? Did the `get_population_extended` funtion change the weights?

kilianvolmer (changes requested):
Thank you for resolving this old issue. The changes look fine, but I would like to see a test for the database conversion. Could you please add that?

pyabc/storage/migrations/versions/2_20260724_add_populations_wall_time.py
arrjon: This is not needed as the other change only adds new database entries, which older versions will just ignore.

test/base/test_storage.py
arrjon: `get_population_extended` doesn't touch the weights, it selects `Particle.w` straight from the DB, so it exposed the raw stored weight. Later `get_distribution` renormalizes weights before they are returned, so this check a different path.

pyabc/visualization/kde.py
arrjon: I think this was commented out at some point because it breaks autoscaling of the y-axis, as this function is repeatedly called. I just removed the dead code. Now I fixed the axis to start at 0, which it should because we are plotting densities.

pyabc/storage/migrate.py
```diff
@@ -14,6 +14,30 @@
 SQLITE_STR = 'sqlite:///'
 
 
+def _alembic_config(db_file: str) -> 'Config':
+    """Create the alembic configuration operating on a database file.
+
+    Parameters
+    ----------
+    db_file: Database file (not URL) the migrations are applied to.
```
kilianvolmer: As we use the URL for most/all other functions, we should maybe allow both here, or add a check and a good error message.

test/migrate/test_migrate.py
```diff
@@ -29,3 +55,215 @@ def test_db_import(script_runner):
...
+def assert_weights_close(actual: dict, expected: dict) -> None:
+    """Assert that two ``{t: {m: weights}}`` dictionaries match."""
+    assert actual.keys() == expected.keys()
+    for t, per_model in expected.items():
+        assert actual[t].keys() == per_model.keys()
+        for m, ws in per_model.items():
+            assert np.allclose(actual[t][m], ws)
+
+
+def to_v1(db_file: str) -> None:
```
kilianvolmer, line 159: Why do we use a custom function for that instead of the migration function?

kilianvolmer: I think it would be preferably to use the migration function plus the drop table.

pyabc/storage/migrate.py
arrjon: Done

test/migrate/test_migrate.py
arrjon: Done

kilianvolmer (approved):
Thank you for implementing all the changes! I think we can merge this now :)

PR #671, opened 2026-07-24
Prepare Release 0.13.0

### LINKED ISSUE #47: Normalization of weights

### LINKED ISSUE #325: Profiling
Add code execution time profiling routines, covering:
* "In-between-iterations" time
* Population size calculation
* Distance function adaptation
* Setting up the parallel pipelines
* Pure simulation time vs. rest

### LINKED ISSUE #667: Walltime plots do not show actual walltime if stored abc runs are resumed
**Bug description**
If a stored abc run is resumed and the `pyabc.visualization.plot_walltime` function is used, then the "walltime" includes the time that has passed between the runs. 

**Expected behavior**
The `plot_walltime` function should only plot the actual walltime of the iterations.

**To reproduce**
Run a fitting, resume it a few hours later and plot the walltime.

Image
Here, the total walltime should be eleven hours instead of 25.

arrjon:
General:
* Add support of python3.14
* Add execution time profiling to the ABC-SMC run, reporting pure simulation time, parallel-pipeline setup, in-between-iterations
  time, and within it the population size calculation and distance function ´
  adaptation. The timings are also returned by ``run_generation``. Resolves #325.
* Minor bug fixes and improved type setting.

Visualization:

* ``plot_walltime`` and ``plot_total_walltime`` now report the actual
  per-generation walltimes and no longer include the idle time that passed
  between a stored analysis and a later resumed run. Resolves #667.

Storage:

* Store the per-generation walltime in the database (new ``wall_time`` column,
  database version 2). Databases created with older pyABC versions must be
  migrated via ``abc-migrate`` before they can be resumed; for such databases
  the walltime plots fall back to the previous, end-time-based behavior.
* Particle weights are now stored using the global normalization (weights sum
  to 1 across all particles of all models, matching the in-memory
  ``Population`` representation). This is bundled into database version 2 and
  handled by ``abc-migrate`` for existing databases. Resolves #47.


arrjon:
I bumped the version to 0.13.0, as we have breaking changes in the database. Ready for the new release now.

.github/workflows/ci.yml
```diff
@@ -26,42 +29,42 @@ jobs:
             python: "3.11"
             toxenv: base
           - os: ubuntu-latest
-            python: "3.11"
+            python: "3.13"
```
kilianvolmer: This choice seems a bit random, being neither the newest nor the oldest version?

doc/conf.py
```diff
@@ -52,6 +55,11 @@
     'sphinxcontrib.jquery',
 ]
 
+autodoc_mock_imports = [
+    "PEtab",
+    "BasiCO"
+]
+
```
kilianvolmer:
```suggestion
```

kilianvolmer: Sorry, this should not have been pushed.

doc/conf.py
```diff
@@ -99,6 +107,27 @@
...
+    if rtd_version and rtd_version not in ("latest", "stable"):
+        return rtd_version
+    try:
+        return subprocess.check_output(
+            ["git", "rev-parse", "--abbrev-ref", "HEAD"],
+            text=True
+        ).strip()
+    except Exception:
+        return "main"
+    return "main"
```
kilianvolmer:
```suggestion
```

kilianvolmer (changes requested):

kilianvolmer (approved):

PR #672, opened 2026-07-27
fix ruff

arrjon:
This pull request modernizes type annotations throughout the codebase by using Python 3.10+ union syntax (`| None`). The main focus is improving code clarity and compatibility with newer Python versions.


pyabc/visualization/model_probabilities.py
```diff
@@ -68,8 +68,8 @@ def plot_model_probabilities_plotly(
     history: History,
     rotation: int = 0,
     title: str = 'Model probabilities',
-    size: tuple = None,
-    fig: 'go.Figure' = None,
+    size: tuple | None = None,
+    fig: 'go.Figure | None' = None,
 ) -> 'go.Figure':
     """Plot model probabilities using plotly."""
     import plotly.graph_objects as go
```
kilianvolmer, line 75: Why do we not import this once for all functions?

kilianvolmer (approved):
Thank you for adding this separate PR!

pyabc/visualization/model_probabilities.py
arrjon: plotly is an optional dependency

